PR #96, opened 2018-03-20
Add featured posts and homepage display

### LINKED ISSUE #89: As a user, I want to see featured updates on the homepage so I can see what's going on at the CDH.
- [x] add most recent 3 featured blog posts to homepage view 
- [x] update homepage template to render a featured post block at the top
- [x] new block should use the post's featured image as a background, or a placeholder
- [x] new block should have pagination control for moving through the featured posts

### LINKED ISSUE #90: As a Content Editor, I want to designate blog posts as featured so I can highlight their importance
- [x] add boolean field to blog post model to indicate if it's featured; default false
- [x] make field editable in the admin interface
- [x] make toggleable in list view

thatbudakguy:
closes #89 
closes #90


cdhweb/resources/views.py
```diff
@@ -44,7 +46,12 @@ class Homepage(TemplateView):
     template_name = 'site_index.html'
 
     def get_context_data(self, *args, **kwargs):
-        # TODO: highlighted/featured blog post
+        # get featured posts if there are any
+        if (BlogPost.objects.featured().exists()):
+            updates = BlogPost.objects.featured()
```
rlskoeser: This _looks_ inefficient to me, although I think it probably actually isn't (because Django is clever).  However, I think it would be cleaner and more readable to do something like:
```
updates = BlogPost.objects.featured()
if not updates.exists():
   updates = BlogPost.objects.recent()[:3]
```
Even if there are featured items, don't we still need `recent()` and `[:3]` ?

In that case, you might do something like this:
```
updates = BlogPost.objects.featured()
if not updates.exists():
   updates = BlogPost.objects.all()
updates = updates.recent()[:3]
```

templates/snippets/carousel.html
```diff
@@ -0,0 +1,41 @@
+{% if updates %}
+<div id="carousel">
+    {% for post in updates %}
+    {% if forloop.first %}
+    <div class="post-update active">
```
rlskoeser: Why not wrap just the `active` attribute in the conditional instead of repeating the whole line?

templates/snippets/carousel.html
```diff
@@ -0,0 +1,41 @@
...
+    {% else %}
+    <div class="post-update">
+    {% endif %}
+        {% if post.featured_image %}
+        <header style="background-image: url('{{ MEDIA_URL }}{{ post.featured_image }}')">
```
rlskoeser: Same comment about conditional attribute.

templates/snippets/carousel.html
```diff
@@ -0,0 +1,41 @@
...
+                <p class="description">{{ post.short_description }}</p>
+            </div>
+        </a>
+    </div>
+    {% endfor %}
+    <div id="post-controls">
+        {% for post in updates %}
+            <a href="#">
+                {% if forloop.first %}
+                <div class="dot active" data-index="{{ forloop.counter0 }}"></div>
```
rlskoeser: Another conditional attribute.

templates/snippets/carousel.html
```diff
@@ -0,0 +1,41 @@
...
+                {% else %}
+                <div class="dot" data-index="{{ forloop.counter0 }}"></div>
+                {% endif %}
+            </a>
+        {% endfor %}
+    </div>
+</div>
+{% else %}
+<div id="spacer"></div>
```
rlskoeser: What's this for?

templates/snippets/carousel.html
```diff
@@ -0,0 +1,41 @@
...
+                <div class="quote">"</div>
+                <div class="heading">Updates</div>
+                <div class="line"></div>
+                <h1 class="title">{{ post.short_title }}</h1>
+                <p class="description">{{ post.short_description }}</p>
+            </div>
+        </a>
+    </div>
+    {% endfor %}
+    <div id="post-controls">
```
rlskoeser: Should this be disabled if there is only one featured post?  I don't know how likely that it is, but seems possible.

cdhweb/blog/migrations/0005_auto_20180319_1212.py
```diff
@@ -0,0 +1,26 @@
...
+
+    dependencies = [
+        ('blog', '0004_blogpost_is_featured'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='blogpost',
+            name='featured_image',
+            field=mezzanine.core.fields.FileField(blank=True, help_text='Appears on the homepage carousel when post is featured.', max_length=255, null=True, verbose_name='Featured Image'),
```
rlskoeser: If you still can, would be nice if you would remove migrations 4 and 5 and regenerate a single one with the nice name that you used for the first one so we don't get an extra unnamed migration just for changes in help text.

cdhweb/blog/models.py
```diff
@@ -37,6 +39,29 @@ def author_list(self):
...
+        '''sort blog posts by publication date'''
+        return self.order_by('-publish_date')
+
+class BlogPostManager(DisplayableManager):
+
+    def get_queryset(self):
+        '''Return default queryset :class:`BlogPostQuerySet`'''
+        return BlogPostQuerySet(self.model, using=self._db)
+
+    def featured(self):
```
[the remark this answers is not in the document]
thatbudakguy: This seems like exactly what we want to avoid duplication. This happens with other models too, so maybe I can make a separate issue dedicated to refactoring for it?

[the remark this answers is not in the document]
rlskoeser: @thatbudakguy great idea - please go ahead and create an issue to document this so we can prioritize it for later work.

PR #97, opened 2018-03-26
Style rich text editor preview; limit available elements

### LINKED ISSUE #94: As a Content Editor, I should only be able to use supported formatting and tags when I edit site content so that the CDH has a uniform web presence.
- [x] remove unstyled tags from TinyMCE
- [x] auto-strip `<span>`s?
- [x] change the Image shortcut/snippet to prompt for `<figcaption>`

### LINKED ISSUE #95: As a Content Editor, I want my rich text editor preview to match the way the content will display on the site so that I don't have to check the published version myself.
- [x] make TinyMCE load portions of our styles that we want

thatbudakguy:
closes #94 
closes #95


thatbudakguy:
Added some more verbose comments to `tinymce_setup.js` and `change_form.html` to indicate why compressor requires us to do it this way.

PR #117, opened 2018-08-21
Revise project pages to display current / past, new page for staff projects

### LINKED ISSUE #98: Factor out manager/QuerySet code duplication
limitations of Mezzanine's `DisplayableManager` have given rise to duplicating methods like `featured()`, `published()`, and `recent()` for  model `QuerySets`. We can use Django's [QuerySet.as_manager()](https://docs.djangoproject.com/en/2.0/topics/db/managers/#create-manager-with-queryset-methods) with a potential `DisplayableQuerysetMixin` to resolve this.

Affected models/managers:
- `BlogPost`
- `Event`
- `Profile`
- `Project`

rlskoeser:
Revises project list view to display current and past projects in separate sections. Adds a new project list view for staff and postdoc projects (also grouped by current and past).

Also fixes #98


cdhweb/resources/models.py
```diff
@@ -81,3 +82,22 @@ class LandingPage(Page, RichText):
...
+    Uses the published check from
+     :class:`mezzanine.core.managers.PublishedManager`.'''
+
+    def published(self, for_user=None):
+        """
+        For non-staff users, return items with a published status and
+        whose publish and expiry dates fall before and after the
+        current date when specified.
+        """
+        if for_user is not None and for_user.is_staff:
```
thatbudakguy: codecov thinks we aren't testing the first case (return all). not sure if that matters.

Issue #110, opened 2018-08-09
As a user, I want to easily read and use the main site navigation so I can get to the content that interests me.


rlskoeser:
## testing notes
zeplin links: [L](https://zpl.io/aMvvknK) / [L expanded](https://zpl.io/bJvvkgW)

- the nav should expand to show "cards" with second and third-level items when you hover over a top-level item
- top-level items should have a larger font size for increased readability
- on a large screen, second-level items should stack horizontally
- on a medium screen/tablet, second-level items should stack vertically
- third-level items should be underneath their second-level parent
- when you scroll down quickly, the popup "cards" should hide themselves along with the nav
- popup behavior should be "generous": the "cards" should stay open as long as you want them to, but get out of your way when you expect them to

## Notes for development
- larger font size
- make hover nav easier to use

jimccasey1:
I played around with the navigation menus on the test site. all looks good, except a couple of qualifications that i moved into a list at the bottom. 

- (yes) the nav should expand to show "cards" with second and third-level items when you hover over a top-level item
- (yes) top-level items should have a larger font size for increased readability
- (yes) third-level items should be underneath their second-level parent
- (yes) when you scroll down quickly, the popup "cards" should hide themselves along with the nav
- (yes) popup behavior should be "generous": the "cards" should stay open as long as you want them to, but get out of your way when you expect them to

Unsure about these: 
- on a large screen, second-level items should stack horizontally 
--> under 'Resources' it looks like the last item 'Course Offerings' wraps around and is a little awkwardly set against the others? otherwise the horizontal stack works on all the second-level items. 
- on a medium screen/tablet, second-level items should stack vertically
--> yes, but I don't have an iPad for testing, so just used Chrome inspector set to screen size of 768x1024.

jimccasey1:
Updated previous comment. All looks good, so closing now.

PR #118, opened 2018-08-22
Feature/nav updates

thatbudakguy:
refs #110 

- updates navigation to use new card style
- adds templates for rendering "cards" with secondary and tertiary navigation
- adds styling for medium-size screens (switches to column stacking)
- changes the trigger for hovering to allow for more time to navigate the menu
- makes the primary navigation font size larger


rlskoeser:
This is generally looking and working really well.   There are a few cases with current content where it doesn't work so well - e.g., 5 top-level menu nav on medium gets bumped down to a second line below the CDH logo, or the long list of Grant subpages on the current Grant section overflows to the right rather than wrapping.  I'm not sure how much we care about that since we don't expect it to be a problem with the revised site architecture, but if it's easy to fix it might be nice to have the flexibility.

Can we get `if page.is_current_or_ascendant` logic and highlight styles on the secondary and tertiary nav templates?

rlskoeser:
@thatbudakguy sounds reasonable to me (regarding highlighting current nav item in secondary/tertiary menus). I approved the PR without requiring the change because I wasn't convinced we needed it. Could be nice to help orient you when you navigate, but not a deal breaker for me.

PR #120, opened 2018-08-23
Feature/revised people pages

rlskoeser:
Revises previous people pages (current/alumni) into new groupings: staff, students (both staff and grantees), and postdocs. Each page has current and alumni sections.

Profile detail page now only displayed for people with `is_staff` flag (probably need to update sitemap logic to make sure they are not included...)

Profile card now displays job title _or_ grant recipient, with dates for non-current positions and grants.  It also links to an external website for people with a link and no local profile.

Added a new `pu_status` field to track grad/undergrad/staff/faculty, populates from ldap. We'll likely want to run init from ldap on accounts en masse to set this.


thatbudakguy:
People are now sorted correctly for me, nice!

I did just notice that the dropdown menu that switches between pages doesn't reflect the status of the current page - e.g. if you switch to "students", the text in the dropdown will still say "staff and affiliates" (even though _inside_ the dropdown, "students" will be highlighted).

Issue #48, opened 2017-09-19
As a user reading a blog post, I want links to next and previous posts so I can browse contents by date.


rlskoeser:
## Notes for testing
Navigate to blog posts and pull up any individual post. At the bottom, you should see links for the next and previous blog post. Check that they display ok, that the link works. Check that the first post doesn't error (only has a previous and no next), check at least one middle post with both next and previous links, and check the last post (no error, only has next and no previous).

rlskoeser:
Came across https://docs.djangoproject.com/en/1.11/ref/models/instances/#django.db.models.Model.get_next_by_FOO and wanted to test it out.  Basically working, but needs testing and design.

@xinyil can you provide suggestion for how to display the next/previous links on the blogpost detail page?  Don't necessarily have to provide a full mock up if you can reference to other styles on the site (basic link style? brief version of the list view with author and title but without opening paragraph?).  I think these should display _after_ any attachments.

norabenedict:
The previous and next links are working properly on the test site. I checked in the middle of the posts for both, and also made sure to look at the first and last post to ensure that only the appropriate links are displayed (and working). 

My only "issue" would be one for design, but it does not effect the properly-working links as they are now. Currently, the "previous" and "next" buttons display the entire title of the articles the come either before or after the blog post that is being viewed. While this might be helpful, I find it problematic for two reasons:
1) At present, the titles are not relegated to a limited space on the page, and instead can span the entire length of the bottom of the page. This is distracting, and also causes a slight formatting issue when the titles are too long (e.g. if a title on the previous page is more than half of the length of the page, it shifts the next page link to below it - making the overall look of these links somewhat messy).
2) I think just a simple "previous" and "next" might be more straightforward for users.

PR #121, opened 2018-08-24
Add next and previous blog post links from blog post detail page #48

rlskoeser:


thatbudakguy:
Tried this out and it looks & works great!

PR #130, opened 2018-08-30
Set up pa11y-ci

rlskoeser:
Set up pa11y-ci to run on travis-ci. Includes a fixture with samples of all basic site content to be loaded and tested.


.pa11yci
```diff
@@ -0,0 +1,10 @@
+{
+    "defaults": {
+        "ignore": [
+            "WCAG2AA.Principle1.Guideline1_4.1_4_3.G18.Fail"
```
thatbudakguy: Do we want to ignore contrast-related errors on cdh-web as well as for PPA? I turned this off because a particular PPA color combination was just below the threshold, but that will probably have changed with the new PPA design, and it might be useful to know if cdh-web has contrast problems.

.pa11yci
```diff
@@ -0,0 +1,10 @@
...
+        ]
+    },
+    "urls": [
+        "http://localhost:8000/"
```
thatbudakguy: We definitely don't need this, as pa11y will hit the homepage already since it's in the sitemap. I tested deleting the "urls" property altogether locally and confirmed it still hit `localhost:8000/`.

package.json
```diff
@@ -0,0 +1,17 @@
...
+  "repository": "https://github.com/Princeton-CDH/cdh-web.git",
+  "author": "The Center for Digital Humanities at Princeton <cdhdevteam@princeton.edu>",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "autoprefixer": "^9.1.3",
+    "postcss-cli": "^6.0.0"
+  },
+  "scripts": {
+    "pa11y": "pa11y-ci --sitemap http://localhost:8000/sitemap.xml"
+  },
```
thatbudakguy: I can use this locally by changing my site domain name to `localhost:8000` in the admin...is there any downside to doing that?

requirements/dev.txt
```diff
@@ -1,6 +1,7 @@
-pytest>=3.6 # required for pytest-django
-pytest-django
-pytest-cov
+# all requirements for production, test, and dev
+-r prod.txt
+-r test.txt
```
thatbudakguy: this is pretty sweet!

thatbudakguy:
@rlskoeser it looks (from travis) like requests made during the pa11y run are completing with 200s for the pages themselves, but 404ing for most of the assets, including our stylesheet. that's kind of concerning since compressor should theoretically already have run, and there could potentially be stuff we're not catching (e.g. contrast issues) related to the styles. does it make any sense to you?

package.json
rlskoeser: I don't understand the question/concern here

thatbudakguy: I just don't know what other things changing the site domain would affect since I don't know much about the sites framework / what things reference it other than the sitemaps. If it does nothing else, great!

thatbudakguy:
found a bunch of these in my logs:
`You're accessing the development server over HTTPS, but it only supports HTTP.`
could it be that pa11y is trying to use HTTPS for some reason?

PR #135, opened 2018-09-05
Feature/speakers

thatbudakguy:
- adds filters for speakers (external to Princeton & are connected to at least one event)
- adds people page displaying list of upcoming and past speaker profiles
- adds logic to display upcoming events relevant to speakers, with link
- adds `show_affiliation` logic to display the institutional affiliation on a profile


cdhweb/people/templates/people/profile_list.html
```diff
@@ -7,7 +7,7 @@
 {% block main %}
 <section class="people">
 
-<h1>{{ title }}</h1>
+<h1>{{ current_title }}</h1>
```
rlskoeser: There will be times of the year that there are no upcoming speakers, so we need some language to display when this section is empty. Maybe we can borrow/adapt from the event list? Could link to main event page (which includes recent past events).
> Next semester's events are being scheduled.  Check back later or view past events.

I think this might be specific to speakers - it seems less likely that the other categories will ever have no current members (maybe possible for faculty affiliates and students?).

cdhweb/people/models.py
```diff
@@ -207,9 +218,9 @@ def _current_grant_query(self):
         )
 
     def current(self):
-        '''Return profiles for users with a current position *or*
-        a current grant, based on start and end dates: either no end date
-        set or an end date in the future.'''
+        '''Return profiles for users with a current position,
+        a current grant, or an upcoming event,
```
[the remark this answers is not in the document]
thatbudakguy: Yep, it's now inaccurate...good catch!

cdhweb/people/templates/people/profile_list.html
thatbudakguy: Adapted the event page language and styled it accordingly.

PR #170, opened 2019-08-26
a11y/main nav kbd

thatbudakguy:
this adds keyboard controls for moving around the main navigation. it borrows from the [w3c nav menubar example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-1/menubar-1.html) where possible.

to use, <kbd>tab</kbd> to the main nav and skip past the cdh icon (home link) until the first item is focused (people). from here, use the below controls to operate the menu.

- <kbd>←</kbd> / <kbd>→</kbd> to move between items/lists
- <kbd>↓</kbd> / <kbd>↑</kbd> to enter/exit a list (open/close a nav card)
- <kbd>home</kbd> / <kbd>end</kbd> to jump to first/last item in a list
- <kbd>esc</kbd> to close the currently active nav card
- <kbd>space</kbd> / <kbd>enter</kbd> to click a currently focused link

note that we currently use the browser default visual focus indicator when navigating with the keyboard. this is so that it doesn't conflict with the active style, so that the user can see the active link (cdh-blue underline) as well as the focused link. it might be worth designing our own focus style to complement the active style in the future.

this solution is implemented using jQuery and definitely duplicates a lot of code - as part of the overhaul of cdh-web we will need to move the frontend pipeline to webpack, at which point we can revisit the code to pull it out into a class, use modern syntax (`let/const`), and remove the jQuery dependency.

### DIFF
--- a/sitemedia/js/local.js
+++ b/sitemedia/js/local.js
@@ -59,51 +59,51 @@ $(document).ready(function () {
     // keybindings for primary nav (people, projects, research, etc.)
     $('.primary-nav > li > a')
         .keydown(function (e) {
-            switch (e.which) {
-                case 35: { // home
+            switch (e.key) {
+                case 'Home': {
                     e.preventDefault()
                     hideCard(e.target.id)
                     var firstCard = $('.primary-nav > li > a').first()
                     firstCard.focus()
                     break
                 }
-                case 36: { // end
+                case 'End': {
                     e.preventDefault()
                     hideCard(e.target.id)
                     var lastCard = $('.primary-nav > li > a').last()
                     lastCard.focus()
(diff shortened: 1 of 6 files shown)


thatbudakguy:
@rlskoeser when you get a second, this is ready for review again.

the problem I discovered is that the homepage was included in `page_branch` but _not_ in `in_menu`, which means that it was secretly the first item in the forloop of links for the primary nav menu despite not being rendered as a link. this meant that the logic I was using to set `tabindex=0` on the first item didn't work, since it was never truly the first item, and so no links actually received `tabindex=0`, making the whole menu unfocusable.

Issue #164, opened 2019-08-21
support customized project page staff lists


thatbudakguy:
we're seeing this issue currently on PPA because the logic for calculating the project team to display on a project page defaults to the current grant. for PPA, this is a dataset curation grant rather than the main sponsored project grant, so it isn't showing the main project team from the earlier sponsored project grant.

the easiest way to resolve this is to add a flag on memberships that can be set to:
- force alum
- force current
- use grant dates (default - will be current if on latest grant)

rlskoeser:
Checked that project page members displayed correctly first before making any changes. I did notice one problem with people showing up as both current members and alumni. Check bitKlavier and Winthrop. It doesn't look like a data issue, and it's only happening with CDH staff who no longer work here - I'm not sure how that could be coming into play, but seems suspicious.


Tested with PPA, setting members of the sponsored project team as current - showed as expected. 
Tested with S&co, setting members of the current grant as alum - showed as expected.

Was nice to be able to edit on the list view. It's a little confusing since you have to know which grant to edit, but fine for now.

rlskoeser:
I looked again and I think the weird behavior I saw *must* have been caused by a data issue (although I'm not certain what or how, in the case of Xinyi showing as both current and alum on bitKlavier). I can't reproduce it in my dev environment, and I can't see anything in the code that would cause it. I refreshed the test site with current production data and everything displays as I expect. And when I added some overrides, those work as expected.

PR #171, opened 2019-08-28
feature/membership flags (#164)

thatbudakguy:
- adds a `status_override` field to `Membership`s that can be set to `past` or `current`
- adds `current()` and `past()` methods to `MembershipQueryset` that include/exclude relevant memberships based on the flag
- removes unneeded methods from `Project`
- adds convenience `alums()` method to `Project` that sorts past memberships by name for display
- adds a `project_membership.html` snippet for repeated code on project detail page
- adds a sample project data fixture
- adapts existing tests

the logic inside `current()` and `past()` on `MembershipQueryset` is a little ugly due to the usage of `Q()` expressions; I tried one or two different ways using `exclude()` and `union()` but they produced unexpected results when trying to sort the returned queryset, and `union()` in particular [makes restrictions](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#union) on what you can do with the queryset afterwards, so I switched to `Q()`.


thatbudakguy:
after moving the main methods for this back to `Project` it was clear that nothing needed their equivalents on `MembershipQueryset`, and indeed according to my IDE nothing else was referencing `MembershipQueryset` at all and it had no other function, so I removed it entirely rather than writing a new set of tests.

PR #175, opened 2019-09-03
Display project grant history summary on project detail page

rlskoeser:
Tweaked your preliminary code slightly
- seemed more logical for default grant sort should be by date oldest to newest, so I changed that and use reverse for display on the template and override in the admin section
- dropped the extra dash (visually redundant since it's right next to a date range)
- added unit tests for the template change

### DIFF
--- a/cdhweb/projects/models.py
+++ b/cdhweb/projects/models.py
@@ -151,11 +151,11 @@ class Grant(DateRange):
     grant_type = models.ForeignKey(GrantType)
 
     class Meta:
-        ordering = ['project', '-start_date']
+        ordering = ['start_date', 'project']
 
     def __str__(self):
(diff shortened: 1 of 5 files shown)


cdhweb/projects/models.py
```diff
@@ -150,9 +150,12 @@ class Grant(DateRange):
...
     grant_type = models.ForeignKey(GrantType)
 
+    class Meta:
+        ordering = ['start_date', 'project']
+
     def __str__(self):
-        return '%s: %s (%s-%s)' % (self.project.title, self.grant_type.grant_type,
-                                   self.start_date.year, self.end_date.year if self.end_date else '')
+        return '%s: %s (%s)' % (self.project.title, self.grant_type.grant_type,
+                                self.years)
```
thatbudakguy: nice, i wondered why this wasn't in use here when i saw it initially.

PR #187, opened 2020-01-06
set up wagtail

### LINKED ISSUE #183: set up wagtail
[basic setup](https://docs.wagtail.io/en/v2.5/getting_started/integrating_into_django.html); [extra params](https://docs.wagtail.io/en/v2.5/advanced_topics/settings.html)
- [x] install wagtail
- [x] update `settings.py`
- [x] update `urls.py`
- [x] add templates for wagtail admin base and login page

thatbudakguy:
closes #183 

adds a few non-essential but useful configs like `TAGGIT_CASE_INSENSITIVE` and `WAGTAIL_USAGE_COUNT_ENABLED` that were suggested in the wagtail setup docs. also ports templates for the wagtail admin backend and login form from mep.

### DIFF
--- a/cdhweb/settings.py
+++ b/cdhweb/settings.py
@@ -108,6 +108,28 @@ RICHTEXT_ALLOWED_TAGS = ('a', 'abbr', 'acronym', 'address', 'area', 'article',
                         'textarea', 'tfoot', 'th', 'thead', 'tr', 'tt', '',
                         'ul', 'var', 'wbr')
 
+####################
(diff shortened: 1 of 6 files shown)


thatbudakguy:
good call on the userbar, thank you! re: the dependencies I think we can see if anything breaks, and if not it should get resolved in 3.1 when we upgrade wagtail to latest.

for the email, I just picked something that sounded right. I guess we could also use whatever address the Django admin uses when it sends emails - is that stored in local settings, or regular settings?

Issue #188, opened 2020-01-22
migration utilities


thatbudakguy:
# dev notes
these methods are needed to enable data migrations for all of the content in the 3.0 release. they are being aggregated as `migration_utils.py`. they are mostly used to mimic wagtail functionality that isn't available in the context of a migration.
## helpers
- [x] `TestMigrations`: helper class for testing migrations 
- [x] `get_parent()`: find parent page of a given page
- [x] `add_child()`: create a new page as the child of a given page
- [x] `create_revision()`: update the content for a page and save it
- [ ] `html_to_streamfield()`: convert raw html string to a `StreamField`
- [ ] `streamfield_to_html()`: reverse of above, used for tests and to reverse migrations
## tests
- [x] `get_parent()` tests
- [ ] `add_child()` tests
- [ ] `create_revision()` tests
- [ ] `html_to_streamfield()` tests
- [ ] `streamfield_to_html()` tests

Issue #180, opened 2019-12-19
As a content editor, I want to create, edit, and delete content pages in wagtail so that I can manage new and existing site content.


thatbudakguy:
# testing notes
**fyi** - for the duration of this big upgrade (v3.0), you'll be able to access the _new_ versions of pages at a special URL, `/pages`. so if you were testing the event co-sponsorship form page, you would go to [/pages/events/cosponsor](https://cdh-test-web.princeton.edu/pages/events/cosponsor/).

you should be able to do a direct comparison by looking at the real site (cdh.princeton.edu) and simply removing the "pages" part of the URL, like [/events/cosponsor](https://cdh-test-web.princeton.edu/pages/events/cosponsor/). and ideally, nothing should appear different at all - we want all the pages to look the same as they did before.

**be careful** because a lot of links (especially in the navigation menus) will still take you to the old versions of pages. if you're testing a page, double-check that the URL has `/pages` in it, because you want to be looking at the new versions.

the big changes will be on the admin side. instead of going to [/admin](https://cdh-test-web.princeton.edu/admin/), you'll now be visiting [/cms](https://cdh-test-web.princeton.edu/cms/). the purpose is the same, but the interface is very different! we're hoping that it will improve lots of workflows for adding and updating content. if there are things you don't like, we might be able to customize them later.

for testing this issue, we want to be able to create new pages on the site. there are two new kinds of pages that can be created: the landing page and the content page. landing pages are ones like `/people` and `/research` that have a big image across the top and a tagline. content pages are basically other "regular" pages that just have text on them - like the about page (`/about`).

content pages:
1. log into the new admin interface (`/cms`)
2. play around with the "pages" tree on the left. if you click the folder icon, you see what pages are listed under that page. if you click the pencil, you edit that page. if you click the arrow, you go down one level in the tree.
3. try creating a new content page. you can create one underneath the homepage, or underneath any of the existing landing pages. you could try a variety of different formatting options for the body text.
4. visit your new content page using its `/pages` url. does everything look right? 

landing pages:
1. repeat steps 1 and 2 above
2. try creating a new landing page. the **only** place you should be allowed to do this is underneath the homepage.
3. visit your new landing page using its `/pages` url. does everything look right? check the header image and tagline, if you set them.

# dev notes
this is a new app that will be created to house page-related parts of resources. eventually it may be fully generalized and so will just be installed and used as-is in other projects.

## new models
- [x] `ContentPage` (port from s&co)
[...]

PR #189, opened 2020-01-28
Chore/migrate pages

thatbudakguy:
this PR serves as a record of work done towards the 3.0 milestone on two issues:
- #188 
- #180 

the migration utilities provided by #188 are a prerequisite to complete all other work in this milestone. view the checklist on #188 for a record of work that has been done up to this point implementing and testing these utilities.

the current focus is a fully-functioning `add_revision` function that can call `html_to_streamfield` to create a page revision with arbitrary html content stored as a `StreamField`. this will allow migrating mezzanine pages to the new content structure, which is the focus of #180.

once this is done, migrations can be written to automatically find and migrate the home page, landing pages, and all content pages (in that order). see [0002_homepage.py](https://github.com/Princeton-CDH/cdh-web/blob/chore/migrate-pages/cdhweb/pages/migrations/0002_homepage.py) for an incomplete example of a migration for the homepage.

in addition to the incomplete functions & tests, one known bug is visible when testing `add_child()`: for some reason the landing page in the fixture is treated as (`instanceof`) `HomePage` in addition to `LandingPage`. it's not obvious why this is the case since the content types in the fixture seem correct.


cdhweb/pages/migration_utils.py
```diff
@@ -0,0 +1,154 @@
...
+from django.core.serializers import serialize
+from django.db import connection
+from django.db.migrations.executor import MigrationExecutor
+from django.test import TestCase
+
+logger = logging.getLogger('wagtail.core')
+
+# see for an explanation of django-treebeard & the `path` attribute:
+# http://www.agilosoftware.com/blog/django-treebard-and-wagtail-page-creation/
+def get_parent(apps, page):
```
rlskoeser, line 14: Why is this necessary? Don't pages have a `get_parent` method?

cdhweb/pages/migration_utils.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # otherwise remove the final 4 digits to get the parent's id (pk)
+    Page = apps.get_model('wagtailcore', 'Page')
+    return Page.objects.get(path=page.path[:-4])
+
+# add_child utility method adapted from:
+# http://www.agilosoftware.com/blog/django-treebard-and-wagtail-page-creation/
+# see also:
+# https://github.com/wagtail/wagtail/blob/stable/2.3.x/wagtail/core/models.py#L442
+def add_child(apps, parent_page, klass, **kwargs):
```
rlskoeser, line 31: Same question about this one as `get_parent`. The docs mention a method, but it doesn't seem to be documented and I don't know if we need a particular version of wagtail before it's available.

cdhweb/pages/migration_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        klass._meta.app_label,
+        klass.__name__,
+        page.url_path
+    )
+
+    return page
+
+# adapted from wagtail's `save_revision`:
+# https://github.com/wagtail/wagtail/blob/stable/2.3.x/wagtail/core/models.py#L631
+def create_revision(apps, page, user=None, created_at=datetime.now(), **kwargs):
```
rlskoeser, line 71: Forgot to ask — what do we need revisions for?  Are they required for content to display?

cdhweb/pages/migrations/0002_homepage.py
```diff
@@ -0,0 +1,74 @@
...
+    site.root_page = home
+    site.root_page_id = home.id
+    site.save()
+
+    # delete the default welcome page
+    welcome = Page.objects.get(title='Welcome to your new Wagtail site!')
+    welcome.delete()
+
+
+def revert_homepage(apps, schema_editor):
```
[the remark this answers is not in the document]
thatbudakguy: Revert logic has to be provided for the `TestMigration` class to work, since it has to know how to rollback the migration it applies in the test. This is also why `streamfield_to_html` exists

cdhweb/pages/models.py
```diff
@@ -0,0 +1,164 @@
...
+from wagtail.images.edit_handlers import ImageChooserPanel
+from wagtail.search import index
+
+from cdhweb.blog.models import BlogPost
+from cdhweb.events.models import Event
+from cdhweb.projects.models import Project
+
+
+#: commonly allowed tags for RichTextBlocks
+RICH_TEXT_TAGS = ['h3', 'h4', 'bold', 'italic', 'link', 'ol', 'ul', 'blockquote']
```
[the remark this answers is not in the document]
thatbudakguy: > We probably need more than this for cdhweb, though

Can you expand on this?

cdhweb/pages/migrations/0002_homepage.py
```diff
@@ -0,0 +1,74 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.27 on 2020-01-08 19:59
+from __future__ import unicode_literals
+
+from django.db import migrations
+from cdhweb.pages.migration_utils import add_child, create_revision
+
+
+def create_homepage(apps, schema_editor):
```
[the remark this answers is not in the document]
thatbudakguy: It definitely opens up a lot of possibilities! I think having a dry run mode for the manage command would probably solve most of what the migration tests are trying to determine; that's a great idea.

cdhweb/pages/models.py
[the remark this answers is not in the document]
rlskoeser: Yeah — I think we might want to do an inventory of the tags that are currently in use and then use then figure out what would be a reasonable set to allow.

Would it make sense for the migration script could have a report mode to summarize what tags are used where? (Maybe some cleanup could be done in advance... 🤔 )

BTW shouldn't this list include `li` ?

cdhweb/pages/migration_utils.py
thatbudakguy: It's not available in the context of a migration.

cdhweb/pages/migration_utils.py
thatbudakguy: revisions associate text content with the page model; it's a one -> many. pages themselves have no content stored in the database and you need at least one revision in order for a page to have content.

thatbudakguy:
Closing, since we can always refer to the conversation on this PR later. I think our eventual approach will not be along the lines of this PR.

Issue #193, opened 2020-08-17
upgrade django + mezzanine


thatbudakguy:
upgrading mezzanine and django together should get us to django 2.x. we will need to use development versions of mezzanine from github.

PR #195, opened 2020-09-01
django/mezzanine upgrades (#193)

thatbudakguy:
this PR:
- upgrades Django to 2.2
- uses the latest available mezzanine (4.3.1) from github
- makes the necessary changes to handle deprecated/changed features in Django
- removes wagtail, which had been previously added in `develop`
- notably changes the location of the media uploads to `/media/` instead of `/static/media`; see DEPLOYNOTES for info


DEPLOYNOTES.rst
```diff
@@ -1,6 +1,14 @@
...
 ============
 
+2.7
+---
+
+- To avoid an ImproperlyConfiguredError in Django 2.2, media must now be stored
+  outside static/. Web servers (Apache) will need to be reconfigured to look 
+  for media in a new location, and a redirect should be created so that links 
+  pointing to /static/media/ are not broken.
+
```
rlskoeser, line 11: I don't think it makes sense to mention the improperly configured error; I think this is already how we should have been doing things and this app just got missed somehow.  Also, unless I've misunderstood, we're not moving where anything is stored, just what urls they are configured under for web access — right? Here's a suggested revision, does this seem clearer to you?

> As part of the upgrade to Django 2.2, media should no longer be served under the url for static site content. Apache should be reconfigured to serve media content at `/media`, and a permanent redirect should be added so that content previously available under `/static/media` will resolve to the new location.

cdhweb/urls.py
```diff
@@ -78,24 +74,10 @@
...
-]
 
-if settings.DEBUG:
-    # serve user-uploaded media in debug mode
-    urlpatterns += [
-        url(r'^media/(?P<path>.*)$', serve, {
-            'document_root': settings.MEDIA_ROOT,
-        }),
-    ]
+]
```
rlskoeser, line 80: Don't we still need a version of this? Looks like we can update to newer syntax https://docs.djangoproject.com/en/3.1/howto/static-files/#serving-files-uploaded-by-a-user-during-development

.gitignore
```diff
@@ -4,7 +4,7 @@ db.sqlite3
 
 # Media ignores
 static
-static/media/
+media
```
rlskoeser, line 7: Surprised this was ever here, actually. Does the media file path in settings need to change, or do we keep that in local settings?  (I don't see it in the diff.)

requirements/prod.txt
```diff
@@ -1,9 +1,9 @@
 ## production requirements
 
-Django>=1.11.15,<2.0
+Django>=2.2,<3.0
 pillow>=5.2.0
-wagtail<=2.3
-mezzanine<=4.3
```
rlskoeser, line 6: Please add a comment documenting briefly what "version" of mezzanine this is (or something). Maybe say Django 2.2 compatible version as of x date ?

requirements/prod.txt
```diff
@@ -1,9 +1,9 @@
...
+git+https://github.com/stephenmcd/mezzanine.git@58aea59#egg=mezzanine
+git+https://github.com/stephenmcd/filebrowser-safe.git@c7eceb0#egg=filebrowser_safe
```
rlskoeser, line 6: If we have to specify this explicitly, probably also worth a brief comment why.

cdhweb/settings.py
```diff
@@ -234,7 +218,7 @@
 # URL that handles the media served from MEDIA_ROOT. Make sure to use a
 # trailing slash.
 # Examples: "http://media.lawrence.com/media/", "http://example.com/media/"
-MEDIA_URL = STATIC_URL + "media/"
```
thatbudakguy, line 237: @rlskoeser here's where the media path changed.

rlskoeser: @thatbudakguy interesting, and that's the only change you had to make? So no need for any deploy or dev notes. We could probably remove the override from the project settings used for deploy after this, but no harm in having it there.

Issue #192, opened 2020-08-03
show staff affiliates on affiliates page


thatbudakguy:
# testing notes
you can see the new affiliates page [on the test site](https://test-web.cdh.princeton.edu/people/affiliates/). Joshua Seufert, a PUL librarian who received a DCG, should be displayed along with Kat Reischl. (in production, we'll need to create a profile for Joshua for this to work, since he doesn't have one yet – but then he will automatically appear).

# dev notes
see [slack conversation](https://cdhprinceton.slack.com/archives/C2VHT183A/p1596485281001400) for context - tl;dr we need to support showing PUL librarians who received DCGs on the affiliates page.

- [x] change "faculty affiliates" page title to "affiliates"
- [x] alter or add queryset methods to `ProfileQuerySet` to support querying for staff with the Project Director role
- [x] ensure that CDH staff and students remain excluded from display on the affiliates page
- [x] change url; add redirect from old url to new (redirectview or migration to add redirect rule, whichever is easier/makes more sense)

relevant portion of `models.py`:

https://github.com/Princeton-CDH/cdh-web/blob/95d2ad778d426eab2d6f1419b91a978f5293da2f/cdhweb/people/models.py#L169-L173

cvansant:
looks good

PR #196, opened 2020-09-02
show staff affiliates on affiliates page #192

thatbudakguy:
see checklist on #192 for a list of changes on this PR.

### DIFF
--- a/cdhweb/people/models.py
+++ b/cdhweb/people/models.py
@@ -166,11 +166,12 @@ class ProfileQuerySet(PublishedQuerySetMixin):
         '''Filter out graduate and undergraduates based on PU status'''
         return self.exclude(pu_status__in=self.student_pu_status)
 
-    def faculty_affiliates(self):
-        '''Faculty affiliates based on PU status and Project Director
-        project role.'''
-        return self.filter(pu_status='fac',
-                           user__membership__role__title='Project Director')
+    def affiliates(self):
+        '''Faculty and staff affiliates based on PU status and Project Director
+        project role. Excludes CDH staff.'''
+        return self.filter(pu_status__in=('fac', 'stf'),
+                           user__membership__role__title='Project Director') \
+                    .exclude(is_staff=True)
 
     def executive_committee(self):
         '''Executive committee members; based on position title.'''
--- a/cdhweb/people/tests.py
+++ b/cdhweb/people/tests.py
@@ -274,21 +274,33 @@ class ProfileQuerySetTest(TestCase):
         assert grad_pi.profile not in Profile.objects.student_affiliates()
         assert grad_pm.profile not in Profile.objects.student_affiliates()
 
-    def test_faculty_affiliates(self):
-        # faculty person who is also project director
-        fac = Person.objects.get(username='Meredith')
-        assert fac.profile in Profile.objects.faculty_affiliates()
+    def test_affiliates(self):
+        # faculty person who is also project director should be affiliate
+        # (josh is already a project director for s&co)
+        faculty = Person.objects.get(username='jk2')
+        assert faculty.profile in Profile.objects.affiliates()
 
-        # non-faculty project director != faculty affiliate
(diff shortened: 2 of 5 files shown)


cdhweb/people/urls.py
```diff
@@ -9,9 +10,10 @@
     url(r'^staff/$', views.StaffListView.as_view(), name='staff'),
     url(r'^postdocs/$', views.PostdocListView.as_view(), name='postdocs'),
     url(r'^students/$', views.StudentListView.as_view(), name='students'),
-    url(r'^faculty/$', views.FacultyListView.as_view(), name='faculty'),
+    url(r'^affiliates/$', views.AffiliateListView.as_view(), name='affiliates'),
     url(r'^speakers/$', views.SpeakerListView.as_view(), name='speakers'),
     url(r'^executive-committee/$', views.ExecListView.as_view(), name='exec-committee'),
-
+    # redirect from /people/faculty -> /people/affiliates
+    url(r'^faculty/$', RedirectView.as_view(url='/people/affiliates', permanent=True)),
```
rlskoeser, line 17: Nice. Should include the trailing slash to avoid redirecting twice.

cdhweb/people/models.py
```diff
@@ -166,11 +166,12 @@ def not_students(self):
...
 
-    def faculty_affiliates(self):
-        '''Faculty affiliates based on PU status and Project Director
-        project role.'''
-        return self.filter(pu_status='fac',
-                           user__membership__role__title='Project Director')
+    def affiliates(self):
+        '''Faculty and staff affiliates based on PU status and Project Director
+        project role. Excludes CDH staff.'''
+        return self.filter(pu_status__in=('fac', 'stf'),
```
rlskoeser, line 172: Might be nice to have a variable for the affiliate statuses, but I guess it may not get us much if we're not using it anywhere else.

cdhweb/people/tests.py
```diff
@@ -419,7 +431,7 @@ def test_staff_list(self):
 
         response = self.client.get(reverse('people:staff'))
         # person should only appear once even if they have multiple positions
-        assert len(response.context['current']) == 1
+        assert response.context['current'].filter(user__username='staff').count() == 1
```
rlskoeser, line 434: Why the username filter? Does the test content now include more than one person? Asking because I can't tell if this is a useful test / if it's actually checking anything.

cdhweb/people/views.py
```diff
@@ -195,17 +195,16 @@ def get_past_profiles(self):
             .order_by('-most_recent')
 
 
-class FacultyListView(ProfileListView):
-    '''Display current and past faculty affiliates'''
-    page_title = 'Faculty Affiliates'
+class AffiliateListView(ProfileListView):
```
rlskoeser, line 198: Does anything need to happen to migrate/convert the existing mezzanine page/placeholder in menus for the current faculty profile page when this change takes effect?

rlskoeser (approved):
Looks pretty good!

cdhweb/people/urls.py
thatbudakguy: good catch, thx

cdhweb/people/tests.py
thatbudakguy: that's exactly why, yeah. a lot of these tests are unfortunately pretty brittle because they're sharing a lot of state via the big fixture...it's not great.

cdhweb/people/views.py
thatbudakguy: just to rename it "affiliates" and edit its text content slightly; I didn't know if that was worth a migration - seems like not.

thatbudakguy: should this be a `DEPLOYNOTES` thing, or is it fine to not address?

rlskoeser: I think useful to document in deploy notes just to make sure we don't forget

PR #203, opened 2020-09-16
Use django-apptemplates to specify richtextpage template inheritance

rlskoeser:
I can't reproduce the bug in my local dev (which I think was always one of the problems with trying to fix this one), but it occurred to me that using `django-apptemplates` might help — it lets us specify the package of the template we want to extend.


thatbudakguy (approved):
for some reason I can't duplicate it in QA now (I thought I was able to before), so I guess let's just merge and test this in prod.

Issue #204, opened 2020-09-16
treat research lead co-PIs position as project director


thatbudakguy:
for geniza we're adding a new "co-PI: Research Lead" position that should function the same as "project director" for the purpose of staff pages. all of the places that specifically look for the "project director" role should look for this one as well.

# testing
anyone with the "co-PI: Research Lead" role on a grant should appear on the affiliates page (test with Marina for Geniza).

thatbudakguy:
confirmed in production that marina is showing on affiliates page.

PR #205, opened 2020-09-16
add research co-PI role as project director (#204)

thatbudakguy:
see #204 

- adds `Role.DIRECTOR_ROLES` so that more than one role can be treated as directorship for the purpose of determining who is a CDH affiliate
- resolves some circular import issues

### DIFF
--- a/cdhweb/people/models.py
+++ b/cdhweb/people/models.py
@@ -12,7 +12,6 @@ from mezzanine.core.models import (CONTENT_STATUS_DRAFT,
 from mezzanine.utils.models import AdminThumbMixin, upload_to
 from taggit.managers import TaggableManager
 
-from cdhweb.projects.models import Role
 from cdhweb.resources.models import (Attachment, DateRange,
                                      PublishedQuerySetMixin)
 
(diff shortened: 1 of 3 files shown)


cdhweb/people/models.py
```diff
@@ -129,7 +130,8 @@ class ProfileQuerySet(PublishedQuerySetMixin):
     student_titles = ['Graduate Fellow', 'Graduate Assistant',
                       'Undergraduate Assistant']
     #: memebership roles that indicate someone is an affiliate
-    project_roles = ['Project Director', 'Project Manager']
+    project_roles = ['Project Director',
+                     'Project Manager', 'Co-PI: Research Lead']
```
rlskoeser, line 134: Not sure that the list of director roles need to be defined on `Role` when we're defining other lists of roles here (esp. since that complicates the imports). Any reason not to just define it here? It doesn't seem to be used in the other file.

rlskoeser (approved):
Looks good, just a question about the best place to define the list of roles for consistency & simplicity.

cdhweb/people/models.py
thatbudakguy: sure, makes sense - that resolves the import issues too

Issue #197, opened 2020-09-02
affiliate grant year ranges on profile list pages are incorrect


thatbudakguy:
this is issue affects the display of grant information on the profiles of current affiliates. because the page tries to display the full range of years that someone has had _any_ grant (for example, S&Co. has had grants for 5+ years), it creates a year range to display.

the issue is that it only displays the _name_ of the most recent grant, so it will write something like `2015-2021 Dataset Curation Grant` which is wrong - the dataset curation grant was only for 2021.

the easiest solution for now is to simply not display the range, so that the information is at least technically correct. thus people who got multiple grants will only display the most recent one, i.e. `2021 Dataset Curation Grant`.

when we have a workable solution to display ranges again, that will be tracked on #222.

for more context, see slack conversation: https://cdhprinceton.slack.com/archives/C2VHT183A/p1599587576088900

# testing notes
- Josh and Kat Reichl are good candidates to test this with, since both of them are currently displaying ranges that are incorrect in production.
- All profiles on the affiliates page in the "current" section should only display the most recent grant (so all dates should be this academic year).
- All profiles on the affiliates page in the "past" section should be unchanged; just give a quick to look to make sure that nothing looks off.

thatbudakguy:
moving out of v2.7 since there's not an easy solution and right now it only affects a few profiles.

thatbudakguy:
if we implement the simplest fix (just display most recent grant), need to make an enhancement to later improve the display again.

cvansant:
@thatbudakguy Looks fine! Only thing is that the link to Reischl's page in Slavic doesn't work. Here is the new link: https://slavic.princeton.edu/people/katherine-hill-reischl

PR #223, opened 2020-11-02
Remove grant year ranges from profile card display

thatbudakguy:
Including grant year ranges created a mismatch between the years
listed and the name of the grant shown, which was only the
most recent one.

See #197

### DIFF
--- a/cdhweb/people/templates/people/snippets/profile_card.html
+++ b/cdhweb/people/templates/people/snippets/profile_card.html
@@ -16,13 +16,7 @@
             {% if show_grant %} {# display grant role #}
                 {% with grant=profile.user.latest_grant %}
                     {% if grant %}
-                    {# NOTE: for unknown reasons, on qa these are not returned as date objects; handle either date or YYYY-MM-DD string#}
-                        <p>{% if profile.first_start and profile.last_end %}
-                        {# NOTE: not using start year because it creates incorrect display values; see issue 197 #}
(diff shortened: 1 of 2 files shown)


cdhweb/people/templates/people/snippets/profile_card.html
```diff
@@ -18,9 +18,10 @@
                     {% if grant %}
                     {# NOTE: for unknown reasons, on qa these are not returned as date objects; handle either date or YYYY-MM-DD string#}
                         <p>{% if profile.first_start and profile.last_end %}
+                        {# NOTE: not using start year because it creates incorrect display values; see issue 197 #}
                             {% firstof profile.first_start.year profile.first_start|slice:":4" as start_year %}
                             {% firstof profile.last_end.year profile.last_end|slice:":4" as end_year %}
-                            {{ start_year }}{% if start_year != end_year %}–{{ end_year }}{% endif %}
+                            {{ end_year }}
```
rlskoeser, line 24: seems a little weird to me not to display the start/end year range for the most recent grant, but I guess this works as a simple, temporary solution

rlskoeser (approved):

cdhweb/people/templates/people/snippets/profile_card.html
thatbudakguy: you're right! it also simplified things a ton to just use start/end for most recent grant. i've changed it. thanks!

Issue #225, opened 2020-11-02
test against postgres in CI


thatbudakguy:
this should be as simple as switching database engines in `local_settings.py`, changing the CI provider to use a different database, and then seeing what breaks. as a bonus, we can use the branch to test locally whether problems like #216 are resolved.

rlskoeser:
Let's test on both postgres and mysql; we want to support both while we're in transition.

Issue #224, opened 2020-11-02
shift CI to github actions


thatbudakguy:
see also #198 

should speed up our build times and be simpler than for other projects, since it doesn't require solr. we could also experiment on this project down the line with automated QA deploys.

PR #228, opened 2020-11-03
GitHub actions CI

thatbudakguy:
Two motivations for doing this now:
- it's faster than travis
- we can test against postgres in CI using github actions instead of travis (#225)

Changes:
- Basic GitHub Actions unit testing workflow
- Use python 3.5.10
- Use oldest avail. mysql in CI (5.6)
- Update python caching strategy in CI
- Use default database port in CI
- Remove unused deploy/config files

refs #224


.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,78 @@
+name: unit tests
+
+on: [push, pull_request]
```
rlskoeser, line 3: Can we add a cron trigger here too? Similar to whatever weekly build we have for travis-ci. And we should turn off the travis-ci integration when we switch over.

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,78 @@
...
+    # We use a service container to avoid needing to set up a local copy of
+    # mysql on the test runner instance. This syntax is similar to the spec of
+    # a docker-compose file. For more, see:
+    # https://docs.github.com/en/free-pro-team@latest/actions/guides/about-service-containers
+    services:
+      db:
+        image: mysql:5.6  # oldest available on docker hub; QA/Prod use 5.5
+        env:
+          MYSQL_DATABASE: cdhweb
+          MYSQL_ROOT_PASSWORD: cdhweb
```
rlskoeser, line 21: I presume these have to match our test settings? Might be nice to set them once for both, but seems fine to hardcode here for now.

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,78 @@
...
+    # We base the python cache on the hash of all requirements files, so that
+    # if any change, the cache is invalidated.
+    - name: Cache pip
+      uses: actions/cache@v2
+      with:
+        path: ~/.cache/pip
+        key: pip-${{ hashFiles('requirements/*.txt') }}
+        restore-keys: |
+          pip-${{ hashFiles('requirements/*.txt') }}
+          pip-
```
rlskoeser, line 66: nice

ci/testsettings.py
```diff
@@ -3,34 +3,34 @@
...
+            'CHARSET': 'utf8',
+            'COLLATION': 'utf8_general_ci'
+        }
     },
 }
 
-# required with django 1.11 when debug is false, even for tests
-ALLOWED_HOSTS = ["*"]
+# required for database access?
+ALLOWED_HOSTS = ['*']
```
rlskoeser, line 27: I don't think this is required for database access, but it does take effect when debug is off. I think it would impact the test client hitting the test server in some of our tests.  Using `*` seems fine for tests, but let's revise the comment.

.github/workflows/unit_tests.yml
rlskoeser: Please create a follow up issue to set up slack notifications for builds (if you haven't already)

rlskoeser (approved):
Looks good to me.

Make sure you update the readme image & link to replace travis with build status when you switch over to this. (Maybe in this PR before merging?)

.github/workflows/unit_tests.yml
thatbudakguy: I'll see if I can make it part of this issue - if it's not trivial, I'll split it out.

.github/workflows/unit_tests.yml
thatbudakguy: They do, yeah - I added a note in the testsettings. Probably if we wanted we could do something slightly smart with parsing the `testsettings.py` file (maybe just `awk` or `sed`?) and set them as envvars earlier in the build, in a similar manner to the `.python-version` trick

ci/testsettings.py
thatbudakguy: For me, no tests at all would pass without this - they all raised
`RuntimeError: Database access not allowed, use the “django_db” mark, or the “db” or “transactional_db” fixtures to enable it.`
So it seems to affect everything. What should the comment say?

.github/workflows/unit_tests.yml
rlskoeser: Environment variables seems simplest. We could just adjust `testsettings.py` to set configurations from environment variables (simpler than parsing and rewriting the file!).

thatbudakguy: Makes sense!

ci/testsettings.py
rlskoeser: That's strange... I wonder if it's because of django's test client, which maybe gets automatically initialized. The error occurred on everything, even `TestClient` tests? 

I think the comment can just say required for tests when debug is false (it isn't database specific, in spite of the errors you're seeing).

thatbudakguy:
@rlskoeser Added README badge, scheduled builds, and a rudimentary slack notification for scheduled builds (you can see the result in `#maintenance`). There are several problems with formatting, and it currently can only post to `#maintenance`, but I think this implementation is enough to merge, and tweaking the slack notification can happen as I develop https://github.com/Princeton-CDH/slack-ci-action.

rlskoeser:
@thatbudakguy updates look good

Issue #217, opened 2020-10-21
As a user, I want to see PGRA profiles listed with students and postdoc profiles listed with staff, so I understand how these people relate to the CDH.


thatbudakguy:
See #218, which operates according to the same logic: display people based on how closely they work with CDH.

- having PGRAs with postdocs implies that they are similarly closely associated with the CDH, which isn't true
- postdocs are much more like staff, and PGRAs are more like students than postdocs

# testing notes
- former PGRAs like Sean Fraga, Jessica Mack should display in the past section of the Students page
- all current and former postdocs should display on the Staff page
- Camey (former PGRA, current postdoc) should display on the Staff page but will also show on the Students page because of being project manager on S&Co

thatbudakguy:
For the postdocs:
- [ ] Remove the postdocs page
- [ ] Remove the `not_postdocs()` constraint on the staff page view
- [ ] Add a redirect from the postdocs page URL to the staff page URL
- [ ] Remove the `postdocs()` and `not_postdocs()` queryset methods

cvansant:
@thatbudakguy This works except that the menu item "postdocs" is still there, but the page isn't. Also idk if I should be a student because of S&Co (though it's fine if you want to do it that way). I was never PM for S&Co as a student (I became PM in July). But I was a regular PGRA in the spring.

thatbudakguy:
thanks for pointing it out - rsk mentioned that there was an easy way to make sure you aren't double-listed, so I tried that and you're now only on the staff page. I also removed the link to the old postdoc page from the menu.

gonna close this but feel free to reopen if it doesn't look right to you.

PR #231, opened 2020-11-04
Remove postdocs page

thatbudakguy:
- Moves postdocs to staff page
- Moves PGRAs to student page
- Removes unneeded ProfileQuerySet methods
- Updates people views
- Adds redirect from postdocs to staff page
- Updates deploynotes
- Updates fixtures & tests

See #217

# notes

I ended up changing `student_affiliates()` to look for _either_ a student PU status or a student title, so that PGRAs could be counted as students without needing to change their PU status (since they aren't students as far as Princeton is concerned).


cdhweb/people/models.py
```diff
@@ -122,15 +122,14 @@ class ProfileQuerySet(PublishedQuerySetMixin):
...
     #: position titles that indicate a person is a postdoc
     postdoc_title = 'Postdoctoral Fellow'
-    #: variant postdoc title for Princeton PGRA
-    postgrad_title = 'Postgraduate Research Associate'
 
     #: position titles that indicate a person is a project director
     director_roles = ['Project Director', 'Co-PI: Research Lead']
 
     #: position titles that indicate a staff person is a student
     student_titles = ['Graduate Fellow', 'Graduate Assistant',
```
rlskoeser, line 130: For a future enhancement, should we put a student flag on the position model? To make it configurable instead of hard-coded here.  (What about UAFs? Do we need that title here too?)

cdhweb/people/tests.py
```diff
@@ -434,7 +411,7 @@ def test_staff_redirect(self):
     def test_staff_list(self):
         # fixture includes staff person with two positions
         staffer = Person.objects.get(username='staff')
-        # postdoc with is_staff should not be listed on staff page
+        # postdoc should not be listed on staff page
```
rlskoeser, line 414: postdoc _should be_ listed?

cdhweb/people/urls.py
```diff
@@ -8,12 +8,13 @@
...
     url(r'^staff/$', views.StaffListView.as_view(), name='staff'),
-    url(r'^postdocs/$', views.PostdocListView.as_view(), name='postdocs'),
     url(r'^students/$', views.StudentListView.as_view(), name='students'),
     url(r'^affiliates/$', views.AffiliateListView.as_view(), name='affiliates'),
     url(r'^speakers/$', views.SpeakerListView.as_view(), name='speakers'),
     url(r'^executive-committee/$', views.ExecListView.as_view(), name='exec-committee'),
     # redirect from /people/faculty -> /people/affiliates
     url(r'^faculty/$', RedirectView.as_view(url='/people/affiliates/', permanent=True)),
+    # redirect from /people/postdocs -> /people/staff
+    url(r'^postdocs/$', RedirectView.as_view(url='/people/staff', permanent=True)),
```
rlskoeser, line 18: Should have a trailing slash on the redirect url

cdhweb/people/views.py
```diff
@@ -125,12 +124,11 @@ class StaffListView(ProfileListView):
...
     def get_queryset(self):
-        # filter to profiles with staff flag set and exclude postdocs
-        # and students
+        # filter to profiles with staff flag set and exclude students
         # (already ordered by job title sort order and then by last name)
-        return super().get_queryset().staff().not_postdocs().not_students()
+        return super().get_queryset().staff().not_students()
         # NOTE: this won't filter correctly if we ever have someone who
-        # goes from a postdoc or student role to a staff position, however
+        # goes from a student role to a staff position, however
```
rlskoeser, line 131: We have an instance of this now, and I'm curious how the code handles it. Is this comment still accurate, and do we care?

Camey was a PGRA and now is a postdoc. Does she show up as a past student because of the PGRA position (which works independent of the student status) and also show up  as staff? Or does she only show up as staff? (I think either one is probably ok, I'm just curious how it works now. But it would be good to update the comment here if it isn't accurate anymore).

rlskoeser (approved):
Changes look good to me. Couple questions about possibly inaccurate comments.

cdhweb/people/tests.py
thatbudakguy: yep, good catch!

cdhweb/people/views.py
thatbudakguy: Glad you mentioned this, because Camey was actually being excluded from the staff page, so the note was correct. I changed it back so that `not_students()` excludes based only on PU Status and updated the note.

Camey now shows on both the current staff page and the current student page, but I think that's because she's a PM on Shakespeare and Company which has a current grant, not because of her past PGRA role.

cdhweb/people/models.py
thatbudakguy: I like that idea; maybe we can move away from using PU status and just determine student/staff listings based on a flag set on the position title. I think UAFs will follow the pattern of already having their status set to Student, so we maybe don't need that here yet, but it'd be easy to add.

Issue #233, opened 2020-11-10
As an admin, I want to associate project contributors with projects based on the dates they were involved instead of by grant, so that I can accurately reflect project team membership over time.


rlskoeser:
## testing notes

Project membership has been revised and now is linked to the project instead of a particular grant, and project memberships have start and end dates of their own distinct from the grants. This update adds those fields, removes the old override status, and auto-calculates start and end dates based on the information in the database. (We expect that some manual correction of calculated dates will be needed afterwards.)

- review project memberships in the admin; you may want to compare with production in some cases
  - check that people who were involved on multiple subsequent grants for a single project now have a single membership entry with combined date ranges with the start of the earliest grant and then end of the most recent grant (examples: project directors on multi-year projects like S&co or PPA)
  - check people who've had multiple roles on a single project have multiple membership records (S&co has some)
  - check that cdh staff who left a project before a grant ended have an earlier end date, based on their position dates (examples: Ben, Xinyi, Jean)
   - check that cdh staff who joined cdh after a project had already started have their cdh start date instead of the grant start date (examples: Nick, Kevin)
   - check that people who had overrides with status "current" in production have no end date set on the corresponding membership (PPA and S&co have some examples)
- check that you can create a new membership; start date should be required now
- check that project pages on the public site display as expected with current team members and alums
- check that affiliate and student pages display correctly

---- 

This change should solve some problems, including specifically #212 

- [x]  add start and end dates to project membership model (both optional for now)
- [x] write a migration to calculate best guess for start and end dates by aggregating memberships by person and role, and using associated grant dates; for CDH staff, use start and end dates as additional limits; remove any extra memberships when multiples are combined into a single range
- [x] create log entries in the migration to document the date calculation
- [x] remove grant and status override fields on project membership
- [x] update project membership admin list and edit views to use the new fields and remove the old ones
- [x] revise project admin page to use the new fields and remove the old ones; if including memberships on project page for large projects is still too slow, remove it
- [x] update project detail page to find and retrieve current and past project members based on dates on the membership instead of the grant
- [x] update impacted affiliate pages to use dates on membership instead of grant to determine whether they are current or past
- [x] write unit tests for date calculations migration

rlskoeser:
@thatbudakguy I think, given the number of pieces this touches plus the addition of log entry and script account, this is a 5. LMK if you disagree.

cvansant:
- Elspeth Green has two S&Co PM entries, and two S&Co researcher entries
[...]

Issue #212, opened 2020-10-19
project admin edit pages load slowly, especially for large projects


thatbudakguy:
## testing notes
Membership has been removed from the project edit page in order to expedite editing for large projects with many members.

Check that the project edit page is more responsive now, and check that you can add and edit memberships separately.

rlskoeser:
This is most evident on S&co, since it has such a large number of grants and team members.

Simplest solution: remove membership inline from main project edit page and edit them directly, not as inlines. 

Next simplest: implement simplest solution + also remove `project` foreignkey from `Membership` model. (It's redundant since projects are currently associated with membership through grants as well as directly). Not sure what the implications are for the code, but expect it wouldn't be too hard to make this refactor (maybe a `project` property to stand in for the removed foreign key).

Complicated solution: revise memberships so they attach to the project directly with start and end dates instead of being linked to specific grants.  This solution is probably out of scope for now, but we might want to create an issue (if we don't already have one?) to track this change to the data model so we can document it. (I know we've discussed it in the past; after thinking about it more I think it would solve a few different problems.)

@thatbudakguy I revised to specify that I was talking about the project admin edit page (in case it was unclear), and relabeled as a bug. Do you agree?

cvansant:
Looks good

PR #234, opened 2020-11-11
Refactor project membership 

rlskoeser:
see #233 and #212 

Now with system user and log entry creation


cdhweb/projects/migrations/0011_calculate_membership_dates.py
```diff
@@ -0,0 +1,108 @@
...
+        if current:
+            # if this membership matches the previous one, extend it
+            if current.user == membership.user and \
+               current.project == membership.project and \
+               current.role == membership.role:
+                # extend the end date
+                current.end_date = calculate_membership_end(membership)
+                # delete the obsolete membership
+                membership.delete()
+                continue
```
thatbudakguy, line 29: there is theoretically a scenario here where a person (maybe a PM?) is involved in the same role with a project that gets two overlapping grants, one of which is longer (a year, say) and the other of which is shorter (a few months) and concludes before the longer one ends. in this case, the end date can get updated to point at the end of the second grant, even though the first one lasts longer.

i have no idea if this would ever happen in practice, but in any case I think you can mitigate it by just adding a check here before extending:
```py
calculate_membership_end(current) < calculate_membership_end(membership)
```

cdhweb/projects/migrations/0011_calculate_membership_dates.py
```diff
@@ -0,0 +1,108 @@
...
+            # set dates based on associated grant & cdh position dates
+            current.start_date = calculate_membership_start(membership)
+            current.end_date = calculate_membership_end(membership)
+
+    # save the last record after loop ends
+    if current:
+        current.save()
+
+    # if any memberships are missing start dates, it will be a problem
+    assert not Membership.objects.filter(start_date__isnull=True).exists()
```
thatbudakguy, line 48: should we just do this at the top of the function? might as well bail if there's nothing to do

thatbudakguy (approved):
gave the migration logic a quick look and i don't see anything amiss. definitely agree this is a 5.

cdhweb/people/models.py
```diff
@@ -100,10 +100,27 @@ def profile_url(self):
...
+            # membership end after grant start
+            # grant start before membership end OR
+            # grant end after membership strt
+            print(mship)
+            print(mship.project.grant_set.all())
+
+            return mship.project.grant_set \
+                .filter(models.Q(start_date__lte=mship.end_date) |
+                        models.Q(end_date__gte=mship.start_date)) \
+                .order_by('-start_date').first()
```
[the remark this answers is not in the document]
rlskoeser: Thanks for catching these print statements and extraneous comments, they were indeed out of date.

I had to revise this anyway to handle memberships with no end dates.

cdhweb/projects/migrations/0011_calculate_membership_dates.py
rlskoeser: Hm. Any idea if this is a real scenario? I don't think our grant dates typically overlap.

I can use a maximum, but the logic is slightly more complicated because in the current logic `None` could be a valid value for an end date if the status override is set to current.

cdhweb/projects/migrations/0011_calculate_membership_dates.py
rlskoeser: Since it's a migration, it should only run when we need it, right after start and end date fields have been created and are null.

cdhweb/projects/migrations/0011_calculate_membership_dates.py
thatbudakguy: It's probably not real, so if it's not trivial to handle let's not worry about it.

thatbudakguy (approved):

PR #242, opened 2020-12-01
Configure and test against postgres in CI

kmcelwee:
Main commits:
* Remove duplicate DATABASES dict 1bfb4e3
* Address posgres-specific contentType bug a61beb5
* Use matrix strategy to test against both mysql and postgres cb0c2b0
* Use real date in lieu of 0 in queryset default Value 9f39fac

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -4,7 +4,7 @@ on:
   push: # run on every push or PR to any branch
   pull_request:
   schedule: # run automatically on main branch each Tuesday at 11am
-    - cron: "0 11 * * 2"
+    - cron: "0 16 * * 2"
 
 env:
   NODE_VERSION: 10
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -38,7 +38,7 @@ jobs:
           MYSQL_DATABASE: ${{ env.DB_NAME }}
           MYSQL_USER: ${{ env.DB_USER }}
           MYSQL_PASSWORD: ${{ env.DB_PASSWORD }}
-          MYSQL_RANDOM_ROOT_PASSWORD: true
+          MYSQL_ROOT_PASSWORD: ${{ env.DB_PASSWORD }}
         ports:
           - 3306:3306
     steps:
@@ -92,6 +92,13 @@ jobs:
             pip-
       - name: Install dependencies
         run: pip install -r requirements/test.txt
+      # For mysql only, grant permission to the DB_USER to create databases so
+      # that it doesn't fail trying to create the django test database
+      - name: Grant permissions to mysql user
+        if: matrix.backend == 'mysql'
+        env:
+          GRANT_CMD: "grant all privileges on *.* to '${{ env.DB_USER }}'@'%';"
+        run: echo "$GRANT_CMD" | mysql -h 127.0.0.1 --user=root --password="$DB_PASSWORD"
       - name: Setup local_settings.py
         run: |
           cp ci/testsettings.py cdhweb/local_settings.py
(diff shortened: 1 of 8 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -1,97 +1,131 @@
 name: unit tests
 
 on:
-  push:     # run on every push or PR to any branch
+  push: # run on every push or PR to any branch
   pull_request:
-  schedule:   # run automatically on main branch each Tuesday at 11am
-  - cron: "0 11 * * 2"
+  schedule: # run automatically on main branch each Tuesday at 11am
+    - cron: "0 11 * * 2"
```
rlskoeser, line 7: not actually 11am, right? (based on this morning's run) should we change either the time or the comment?

thatbudakguy:
fyi for @rlskoeser: we traced the strange error in the tests to the `get_past_profiles()` method of the `StudentListView`, where we're doing a `Max()` on membership dates that needs a default value in case of null:
https://github.com/Princeton-CDH/cdh-web/blob/f43a656d77bff9559b1de02fe0a9e71f5977fa61/cdhweb/people/views.py#L152-L174
The default was `'0'`, but postgres seemed to require a "real" date (we tried with literal 0 instead of string and it got turned into a string by the ORM and still failed). the current solution is to just pass a really early date, which seems to work for mysql as well. If we don't care about mysql compatibility, we could presumably just remove the default entirely, since the comment indicates that needing a default is specific to mysql.

.github/workflows/unit_tests.yml
```diff
@@ -1,97 +1,131 @@
...
-      uses: rtCamp/action-slack-notify@master
-      if: ${{ github.event_name == 'schedule' || github.event_name == 'push' }}
-      env:
-        SLACK_WEBHOOK: ${{ secrets.ACTIONS_SLACK_WEBHOOK }}
-        SLACK_CHANNEL: cdhweb
-        SLACK_TITLE: "Run #${{ github.run_number }}> for workflow `${{ github.workflow }}`: ${{ job.status }}"
-        SLACK_MESSAGE: "<https://github.com/${{ github.repository }}/|${{ github.repository }}@${{ github.ref }}>"
-        SLACK_FOOTER: "<https://github.com/${{ github.repository }}/commit/${{ github.sha }}|View commit>"
-        MSG_MINIMAL: true   # use compact slack message format
+      # Set the value of DJANGO_DB_BACKEND which is used in ci/testsettings.py to
```
rlskoeser, line 45: Why does everything look different here? Did the indentation change? It looks like it is substantially the same; did anything change I should review?

ci/testsettings.py
```diff
@@ -7,19 +7,12 @@
 # .github/workflow .yml files.
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.mysql',
-        'USER': 'root',
-        'NAME': os.getenv('MYSQL_DATABASE'),
-        'PASSWORD': os.getenv('MYSQL_ROOT_PASSWORD'),
+        'ENGINE': 'django.db.backends.%s' % os.getenv('DJANGO_DB_BACKEND'),
```
rlskoeser, line 10: nice

requirements/test.txt
```diff
@@ -6,3 +6,4 @@ pytest>=4.6 # required for pytest-cov
 pytest-django
 pytest-cov
 wheel
+mysqlclient # tested against in CI for now
```
rlskoeser, line 9: Will we need this for getting 2.8 mysql production synced to a 2.8.1 postgres production, or are the expected changes for 2.8.1 minimal enough we won't need to deploy it to the old production server before migrating?

(otherwise, seems like a great solution)

cdhweb/people/views.py
```diff
@@ -163,12 +163,12 @@ def get_past_profiles(self):
                 Case(
                     When(user__membership__end_date__isnull=False,
                          then=Max('user__membership__end_date')),
-                    default=Value('0')
+                    default=Value('1900-01-01')
```
rlskoeser, line 166: aha, nice

rlskoeser (approved):
Looks good!

.github/workflows/unit_tests.yml
thatbudakguy: yes, as we discovered! just to double-check...if it ran at 6am, which is 5hrs earlier than scheduled, then we need to shift the cron forward 5hrs, which would be 4pm or 16, so `0 16 * * 2`. does that sound right?

.github/workflows/unit_tests.yml
thatbudakguy: several things did change; sorry git isn't highlighting them very well! the main thing to notice is lines 95-101; this tripped me up so I wrote a comment explaining why it's necessary (for mysql only). other than that, I tinkered with the format of the slack message a little, but there aren't substantial changes below this line.

cdhweb/people/views.py
thatbudakguy: @kmcelwee's idea! sometimes the simplest solution is best, and it works on mysql too, which is nice.

requirements/test.txt
thatbudakguy: I think the only other change for 2.8.1 is a bug that's a one-line change (have solution locally), so seems like the latter.

.github/workflows/unit_tests.yml
rlskoeser: ok, I see it now. Thanks for pointing it out.

PR #258, opened 2020-12-14
Automatically publish dbdocs with github actions

kmcelwee:
Two questions :

* Dbdocs requires global installation with npm. Without the global flag, it fails. I wrote it out, but I assume there might be a better way to do this?
* Should I keep a copy of `cdh-web.dbml` in the repo? That dbml will not be used for publishing to dbdocs, and I can see it being confusing.


thatbudakguy:
re: installing globally, you could try installing normally instead (don't use `-g` flag) and then replacing calls to `dbdocs` with `npx dbdocs`, which should run the locally-installed copy of it (see [`npx`](https://www.npmjs.com/package/npx)). might not work, but worth a shot.

if that doesn't work, we can try implementing the commands as [npm scripts](https://docs.npmjs.com/cli/v6/using-npm/scripts), so that you could call them with e.g. `npm run dbdocs`, which will automatically run the locally-installed version.

requirements/prod.txt
```diff
@@ -16,3 +16,4 @@ icalendar
 attrdict
 django-fullurl
 django-apptemplates
+django-dbml
```
thatbudakguy, line 19: am i wrong in thinking this is a dev or test requirement rather than prod? it only ever runs in CI or locally, right?

thatbudakguy (approved):
looks good - re: your comment about keeping the dbml file around, I think we should `.gitignore` it.

requirements/prod.txt
rlskoeser: Agree that it should be a dev requirment.

cdhweb/settings.py
```diff
@@ -323,6 +323,7 @@
     "fullurl",
     "django_cas_ng",
     "pucas",
+    "django_dbml",
```
rlskoeser, line 326: When you switch this to a dev requirement, you can't include it here — there should be some examples at the end of the settings file where dependencies are imported and added to installed apps when present.

README.rst
```diff
@@ -109,6 +109,11 @@ directory::
 When building documentation for a production release, use `make docs` to
 update the published documentation on GitHub Pages.
 
+On every commit, GitHub Actions will generate and then publish a database diagram to `dbdocs @ princetoncdh/cdh-web <https://dbdocs.io/princetoncdh/cdh-web>`_. But to generate locally, install and log into dbdocs. Then run::
+
+    python manage.py dbml > cdh-web.dbml
+    dbdocs build cdh-web.dbml --project cdh-web
+
```
rlskoeser, line 116: Thanks for documenting

.github/workflows/unit_tests.yml
```diff
@@ -109,6 +109,15 @@ jobs:
...
         uses: codecov/codecov-action@v1
+        # Build and publish dbdocs
+      - name: Download dbdocs
+        run: npm install -g dbdocs
+      - name: Generate DBML
+        run: python manage.py dbml > cdh-web.dbml
+      - name: Login to dbdocs
+        run: echo ${{ secrets.DBDOCS_KEY }} | dbdocs login
+      - name: Push to dbdocs
+        run: dbdocs build cdh-web.dbml --project cdh-web
```
rlskoeser, line 120: Should this be part of unit tests?

I think a separate github action would make more sense; we don't need to regenerate these all the time. Suggest only run on push to `main`.

rlskoeser, line 120: Can we do `cdhweb` instead of `cdh-web` for the project name? Nick and I just refactored the deploy scripts to use the name without the hyphen, we both find it more readable.

rlskoeser (commented):
Looks good. Would like to see it as a separate github action; I think it only needs to run on pushes to `main`.

I don't think we should check the dbml file into git.

Can we add a build status badge to the readme that links to the dbdocs?

.github/workflows/unit_tests.yml
kmcelwee: ✅ addressed with fad2128

kmcelwee:
@rlskoeser @thatbudakguy thanks for your help! Updates since your review:

* Remove dbml and add dbml output to gitignore 6bd9f5e 2193bf4
* Change from cdh-web to cdhweb fad2128
* Separate dbdocs into its own workflow
* Shift django-dbml to test requirements and update settings 453ecc4
* Remove global installation of dbdocs and update to using npx f7282a0 36761d8
* Add dbdocs badge to readme 700987a

Let me know if you'd like to look it over again before I merge!

cdhweb/settings.py
```diff
@@ -398,6 +397,7 @@
     "debug_toolbar",
     "django_extensions",
     "compressor",
+    "django_dbml",
```
rlskoeser, line 400: FYI, this will fail once we remove mezzanine

.github/workflows/dbdocs.yml
```diff
@@ -0,0 +1,61 @@
...
+    services:
+      postgres:
+        image: postgres:12
+        env:
+          POSTGRES_DB: ${{ env.DB_NAME }}
+          POSTGRES_USER: ${{ env.DB_USER }}
+          POSTGRES_PASSWORD: ${{ env.DB_PASSWORD }}
+        ports:
+          - 5432:5432
+    name: Python unit tests
```
rlskoeser, line 24: Is all the database setup required? I guess the script won't run without a valid config even though it doesn't actually use the db?

rlskoeser:
@kmcelwee updates look good! My only concern is the optional apps thing — but I think I'm willing to live with that temporarily in favor of getting this merged.

.github/workflows/dbdocs.yml
kmcelwee: https://github.com/Princeton-CDH/cdh-web/runs/1559007185?check_suite_focus=true
Yeah! I tried removing it but it caused errors. But as I look again, I'm now thinking though that it's because I just didn't have the environment variable defined. I'll try dropping the postgres service and just running with:
```
- name: Set django database backend adapter
   run: echo "DJANGO_DB_BACKEND=postgresql" >> $GITHUB_ENV
```

Issue #240, opened 2020-11-30
set up visual regression testing


rlskoeser:
- [x] install percy and dependencies
- [x] create python/selenium script to run test server and take DOM snapshots to send to percy
- [x] add github actions workflow for running script
- [x] add test site fixture function based on new pytest fixtures; clean up old/unused fixtures
- [x] set up slack notifications for percy
- [x] update README

thatbudakguy:
see https://github.com/Princeton-CDH/cdh-web/pull/260

PR #260, opened 2020-12-15
Set up visual regression testing with Percy

thatbudakguy:
- Install JS and Python dependencies for running Percy
- Create testing script and npm command for running in CI
- Add github actions workflow for running visual tests
- Refactor pytest fixtures to factory functions so they can be used to generate test data
- Create management command for populating test site
- Make pytest fixtures always use constant time (e.g. not `timezone.now()`) for consistent visual output
- Add Percy-specific CSS so that paid fonts can be loaded remotely from CDH site in CI

See #240

just a little more info on how this works in CI:
- the `make_test_site` management command calls a whole bunch of functions that populate the test site with data, using the same logic as our pytest fixtures (since pytest fixtures can't be called directly)
- we run django-compressor once to generate all the static files
- calling `npm test` uses the `start-server-and-test` package to run `npm run serve`, which evaluates to `python manage.py runserver --insecure` (insecure needed to serve static files with debug off)
- once http://localhost:8000 responds to an HTTP request (server is ready), `start-server-and-test` calls `npm run percy`, which evaluates to `percy exec -- python ci/visual_tests.py` (`percy exec` is a wrapper for running python scripts that makes using the percy API easier)
- the `ci/visual_test.py` script runs, which visits several different pages and takes DOM snapshots of them (this is a "bundle" of rendered HTML and CSS, but it runs with javascript off, and is independent of the browser used. we use chrome)
- the snapshots are uploaded to percy servers in the cloud, where they are rendered in parallel using our configured browsers (can be set in the web interface for percy) and configured device/screen widths (can be set in the `.percy.yml` config file)
- when finished, the rendered images are stored and compared to the previous baseline. diffs are highlighted and any diffs outside the configured tolerance (can be changed) will trigger a pull request annotation saying the changes need review for the PR to pass
- accepting the changes in percy's web UI will set a new baseline for the next diff.


.github/workflows/visual_tests.yml
```diff
@@ -0,0 +1,87 @@
+name: visual tests
+
+on:
+  push:
+    branches: [main, develop]
+  pull_request:
+    branches: [main, develop]
```
rlskoeser, line 7: do we need it on main? what's the use case for that, hotfix release?

rlskoeser: Is there a way to filter the commit path so we only run this on commits that impact the templates? (Although I guess that might be hard to determine, since it wouldn't just be templates since view or even model changes could potentially impact the html.) Asking because I saw something about usage limits somewhere.

ci/visual_tests.py
```diff
@@ -0,0 +1,85 @@
...
+    percy_snapshot(browser, "Blog List Page")
+
+    # blog post
+    browser.get("http://localhost:8000/updates/2021/07/06/a-big-announcement/")
+    percy_snapshot(browser, "Blog Post")
+
+    # 404 page
+    browser.get("http://localhost:8000/bad-url")
+    percy_snapshot(browser, "404 Page")
+
```
rlskoeser, line 79: Hmm, might be an argument for adding a url to let us trigger the 500 page (or any other supported error pages) so we can include it in visual testing. Maybe we can add it as a test url or something, so it isn't always there in production.

ci/visual_tests.py
```diff
@@ -0,0 +1,85 @@
...
+    options.add_argument("--headless")
+    return webdriver.Chrome(options=options)
+
+
+def run():
+    """Take DOM snapshots of a set of URLs and upload to Percy."""
+    browser = setup()
+
+    # homepage
+    browser.get("http://localhost:8000/")
```
rlskoeser, line 29: Would it make sense to have a list of urls that should be tested? (Or I suppose a dict with url and name). It seems like that would be easier and more readable to manage, although I'm not sure how you'd handle the urls that require the toggle click interaction — maybe that's could be a separate list, since they are all handled the same way (and would probably all change when/if we revise the nav filter logic).

.github/workflows/visual_tests.yml
```diff
@@ -0,0 +1,87 @@
...
+        run: python manage.py migrate
+
+      - name: Load test site data
+        run: python manage.py create_test_site
+
+      - name: Build static assets
+        run: python manage.py compress
+
+      - name: Take DOM snapshots and upload to Percy
+        run: npm test
```
rlskoeser, line 85: I was confused about this at first (i.e., why are we running all the npm tests) until I remembered that we don't have any existing npm tests for cdhweb.  Could we come up with a standard name for this that we can use across projects? Presumably we wouldn't run all the npm tests when we're doing the percy testing on an application that does have npm tests.

.percy.yml
```diff
@@ -0,0 +1,17 @@
+version: 2
+snapshot:
+  widths:
+    - 375
+    - 768
+    - 1280
+  minHeight: 1024
+discovery:
+  allowedHostnames: []
+  networkIdleTimeout: 100
```
rlskoeser, line 10: Is this configured according to our breakpoints? (I'm wondering how much it will need to change for other sites)

rlskoeser (approved):
This is really cool, and exciting to see it all set up and running. Looking forward to figuring out the workflow so we can start using it.

Do we need additional team documentation somewhere about percy accounts/organization, percy token? Can we just generate our own percy token if we need to run it, or do we need to use a shared one ?

.github/workflows/visual_tests.yml
thatbudakguy: i was wondering the same about `main`! from percy docs it appeared that the standard practice was to leave it on, but set percy to "auto-approve" the build, which I don't really understand. I'm fine with only doing `develop` for now.

re: limiting to paths, we could take the opposite approach and _exclude_ some paths (docs, etc), but generally when touching these files I just put the magic `[ci skip]` in the commit message so that it skips the entire gh actions build.

ci/visual_tests.py
thatbudakguy: yeah, agree — we've had that once or twice while testing and it seems useful here for sure. I can add back the super simple version that I had before, which just maps `/500` to a view function that divides by zero 😆

.github/workflows/visual_tests.yml
rlskoeser: ooh, I wasn't aware of that particular magic incantation

I was thinking more of the case where you want tests to run but don't need visual testing. Probably not something we need right now!  We can always tune it later, especially if the usage becomes a problem.

ci/visual_tests.py
thatbudakguy: i had a list originally (although dict makes sense to cover the titles too) but got rid of it for that reason (the ones that require extra interaction). i guess the question is if we expect to add any other kinds of interaction on any pages — for e.g. PPA I can see testing _lots_ of interactions, especially on the search page, so it can get complicated (and specific to an individual page) quickly.

one thing i could do here is just always try to "click" the dropdown even if the page doesn't have one — i don't think selenium will error, it just won't do anything for that particular page. so that way there only needs to be one "visit" function that covers all the pages

.github/workflows/visual_tests.yml
thatbudakguy: that's a good idea — this config (using `npm test`) is copied from percy docs; in JS conventions the `npm test` command actually isn't guaranteed to do any particular thing at all, so it would be good to establish our own convention instead. maybe `npm test:visual` or something

.percy.yml
thatbudakguy: yeah, it uses the values from cdhweb sass sources. fwiw i think these breakpoints are different from most other projects, but the others are all the same — at least ppa and s&co definitely are; not sure about derrida

thatbudakguy:
re: percy accounts/organization, I'm not sure about documentation...I think our only actions with that will be adding/removing user accounts, which is all done through browserstack. it seems similar to adding/removing people from github or creating repos, which isn't really a thing we document right now.

about the tokens, those are tied to a particular project. as we add new repos to percy, we can copy the token for that project into the repo's secrets on github (not org-wide secrets) and then that's all you need to do to make it available in CI. if you want to run a percy build locally, you can look up the token on percy's web UI and do `export PERCY_TOKEN=dsljasou394820u` locally before running the `npm test` command.

.github/workflows/visual_tests.yml
thatbudakguy: yeah, it's a convention i think from travis that github actions also adopted. the [allowed magic words are described in this post](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/).

cdhweb/pages/management/commands/create_test_site.py
```diff
@@ -0,0 +1,61 @@
...
+    make_events,
+    make_blog_link_page,
+    make_blog_posts,
+)
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        """Fill the database with models to approximate a real site."""
+        # NOTE not idempotent! running more than once will duplicate models.
```
kmcelwee, line 24: Would you mind writing up a section of the readme to document all this?

cdhweb/pages/tests/test_templates.py
```diff
@@ -166,80 +174,79 @@ def test_attachments(self, client, content_page, attachment):
...
+            {
+                "image": test_svg,
+                "alternative_text": alt_text,
+                "caption": caption,
+                "extended_description": desc,
+            }
+        )
+        assert ("<figcaption>%s</figcaption" % caption) in html
         assert '<div class="sr-only" id="graphsvg-desc">' in html
         assert desc in html
```
kmcelwee, line 252: I assume you just ended up applying black to this file?

ci/visual_tests.py
kmcelwee: Or just two loops, one that requires an interaction and one that doesn't?

kmcelwee (approved):
Yay 🎉 thanks for figuring this all out. 

Having a documentation would be great for this. I feel like maybe https://princeton-cdh.github.io/ would be a good place for that, and then link to it in the readme? It'd be great to have a "recipes" section in the CDH-dev site for things like the black pre-commit and percy that we'll want to embed in multiple projects.

ci/visual_tests.py
thatbudakguy: I think I'm going to leave this as it is for now, since we will soon write tests for a page (or pages) that will likely have new types of interaction not covered here — the search page. I don't think we know enough yet about how best to refactor this, and it may continue to get complicated as we test more behaviors.

cdhweb/pages/tests/test_templates.py
thatbudakguy: Yeah, sorry about all the noise...I apply it habitually to files I touch now, and forgot it wasn't on cdhweb yet!

PR #261, opened 2020-12-17
Chore/landing page template

kmcelwee:
Hey @thatbudakguy, I ended up removing `|richtext_filters` from the content / body. Is there a wagtail replacement?


thatbudakguy:
@kmcelwee good question; it seems like not - I went back and looked at some of the wagtail templates for s&co, and it looks like they generally follow the same format, which you can see [in this snippet](https://github.com/Princeton-CDH/mep-django/blob/main/mep/pages/templates/pages/snippets/body.html).

in mezzanine, you generally have a single field containing all the "rich text" for the page, which you output in the template and run through the "rich text filters" (controlled by [`RICHTEXT_FILTERS`](http://mezzanine.jupo.org/docs/configuration.html#richtext-filters)) which define the transformations you want to do on it.

in wagtail, you instead have a series of "blocks" that the user inserted in the editor, which can hold anything (images, text, etc.) and are of different types. so, we use the idiom you can see in the s&co link above (and on the new homepage template for cdhweb) of:
1. creating a container `<div>` (usually with the class `richtext`)
2. iterating over all of the `blocks` in `page.body`
3. calling `include_block` on each block, which is a django template tag available in `wagtailcore_tags`.

you'll want to convert the new landingpage template to use this idiom, instead of accessing `page.landingpage` (which will no longer exist, i think). you can ignore the `formatcitation` filter in the s&co example; that's specific to s&co.

thatbudakguy (changes requested):
i'd like to have template tests on this branch as well before merging the new template. let me know if you want to pair on porting the existing view tests.

thatbudakguy:
moved to #263

Issue #244, opened 2020-12-08
As a content editor, I want existing content pages migrated from mezzanine to wagtail so that I can manage them in the new system.


thatbudakguy:
# testing notes
see the **fyi** section on #180 before testing this, if you haven't read it yet.

as part of the v3.0 upgrade, we are using a script to convert all of the old pages (the ones in `/admin`) into their new counterparts (the ones at `/pages` URLs, which are editable in `/cms`). we're calling this process the "exodus", because it isn't quite a migration. this issue (and others like it) are for testing that certain kinds of pages were "exodized" correctly - that the new versions got created and work just like the old ones (or better!)

for this issue, we need to check that three types of pages were correctly "exodized" and are available under `/pages`:
- the homepage (`/pages/`)
- all the landing pages (`/pages/people`, `/pages/research`, `/pages/projects/about`, `/pages/engage`)
- all the content pages (too many to list here)

notice that `/events` and `/updates` are not landing pages; they're special and will be tested on #252 and #250 respectively. also, `/projects/about` is special - we wanted `/projects/` to just be a list of projects, so the actual landing page is at `/projects/about`. make sure you go to `pages/projects/about` to test it.

- for the homepage and landing pages, load up the current site (cdh.princeton.edu) and just do a quick comparison to make sure nothing looks out of place in the new counterpart.
- for the content pages, you can explore the pages tree in wagtail and just pick a few to examine; anything under `/engage` is a safe bet. stay away from `/projects`, `/events`, `/people`, and `/updates` – most of those special pages haven't been "exodized" yet, so you'll see a lot of blank pages. we'll be testing them later.
- make sure to also test the two embedded forms `/pages/events/cosponsor` and `/pages/engage/consult`; these are sorta "special" but we made them part of this issue, so they should work.

# dev notes
depends on #180 

this will be a script that can be run by a manage command that uses the tools prototyped in #243 to query for all the existing content page (content page, landing page, homepage) and create their new wagtail versions from #180.

likely order is homepage first, then landing pages, then content pages. probably we will have logic like "if this page's parent hasn't been migrated, do that first, then create it as a child of the parent". maybe we can do something clever and request a flattened page tree from mezzanine so that we do it in top-down order.

- [x] migrate homepage
- [x] migrate landing pages
- [x] migrate content pages (recursively)
- [x] ensure page visibility status (draft/published) is migrated
- [x] generate simple report of pages with keywords so data is not lost (but do not migrate)
[...]

PR #263, opened 2020-12-18
Chore/migrate pages

kmcelwee:
Addresses #180 and #244.


cdhweb/pages/admin.py
```diff
@@ -0,0 +1,13 @@
...
+from wagtail.images.models import Image
+
+# unregister wagtail content from django admin to avoid
+# editing something in the wrong place and potentially causing
+# problems
+
+admin.site.unregister(Page)
+admin.site.unregister(Site)
+admin.site.unregister(Image)
+admin.site.unregister(Document)
```
rlskoeser, line 13: Is this necessary? Does wagtail register them by default?

cdhweb/pages/apps.py
```diff
@@ -0,0 +1,9 @@
+from django.apps import AppConfig
+
+
+class PagesConfig(AppConfig):
+    name = 'cdhweb.pages'
+    # NOTE we have to relabel this to avoid a conflict with mezzanine's pages;
```
rlskoeser, line 6: thanks for this note; should we also note that template directory must match this? (in case we ever want to change it later)

cdhweb/pages/models.py
```diff
@@ -0,0 +1,171 @@
...
+]
+
+
+class BodyContentBlock(StreamBlock):
+    '''Common set of blocks available in StreamFields for body text.'''
+    paragraph = RichTextBlock(features=PARAGRAPH_FEATURES)
+    image = ImageChooserBlock()
+    document = DocumentChooserBlock()
+    embed = EmbedBlock()
+    migrated = RichTextBlock(
```
rlskoeser, line 34: should we have a comment on this one?

cdhweb/pages/wagtail_hooks.py
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    features.register_editor_plugin(
+        'draftail', feature_name, draftail_features.BlockFeature(control)
+    )
+
+    features.register_converter_rule('contentstate', feature_name, {
+        'from_database_format': {tag: BlockElementHandler(type_)},
+        'to_database_format': {'block_map': {type_: tag}},
+    })
```
rlskoeser, line 34: no longer needed — wagtail supports block quotes now without this

cdhweb/urls.py
```diff
@@ -84,12 +84,12 @@
...
-        import debug_toolbar
-        urlpatterns += path("__debug__/", include(debug_toolbar.urls)),
-    except ImportError:
-        pass
+# if settings.DEBUG:
+#     try:
+#         import debug_toolbar
+#         urlpatterns += path("__debug__/", include(debug_toolbar.urls)),
+#     except ImportError:
+#         pass
```
rlskoeser, line 92: did we determine whether or not we need this?

templates/base.html
```diff
@@ -83,9 +83,11 @@
     <footer>
         {% include 'snippets/subscribe.html' %}
 
+        {% comment 're-add footer menu when exodized' %}
```
rlskoeser, line 86: works for me, thanks for noting

rlskoeser (approved):
Tests look good. Few small comments/questions

cdhweb/pages/admin.py
thatbudakguy: I checked and it does (although I don't see `Page` here, so I guess we can leave that one out).

Screen Shot 2020-12-21 at 2 50 22 PM

thatbudakguy: Looks like it's just documents and images, actually. But I have no idea where Comments comes from...is that a Mezzanine thing?

rlskoeser: Not sure! "generic" isn't very helpful, is it. I don't see any mention of it in django docs; I think mezzanine has some commenting support but I don't think we would have turned it on.

rlskoeser: oh - it looks like it's a mezzanine generic http://mezzanine.jupo.org/docs/_modules/mezzanine/generic/models.html

thatbudakguy: ok - I'll leave it alone since it'll go away when mezzanine does, then.

Issue #267, opened 2020-12-23
As a moderator, I want People automatically organized into person list pages so I don't have to manually manage these pages.


thatbudakguy:
# testing notes
This issue refers to the person list views on the current CDH site - [sponsored projects](https://cdh.princeton.edu/projects/), [staff projects](https://cdh.princeton.edu/projects/staff/), and [working groups](https://cdh.princeton.edu/projects/working-groups/) (we treat working groups as a type of project for simplicity).

## as a user
We want to make sure that the list pages exist on the new site at the same URLs, and that the same people appear on each list. The corresponding URLs on the test site are:

- <https://cdh-test-web.princeton.edu/people/staff/>
- <https://cdh-test-web.princeton.edu/people/students/>
- <https://cdh-test-web.princeton.edu/people/affiliates/>
- <https://cdh-test-web.princeton.edu/people/executive-committee/>

check a few people on each list and confirm that things look the same as they do in production (the real cdh site). there may be small differences in the images displayed on people's "cards"; this is an acceptable difference (unless they look totally horrendous). for people, it's important to also confirm that the right text is shown underneath the person's name, since this **changes based on which list you're looking at**.

the "students" and "affiliates" pages have page intros (the short bits of text at the top of the list); you should see this same text on the test site versions of those pages.

## as a moderator
In the [admin site](https://cdh-test-web.princeton.edu/cms/), expand the main page tree under "home", expand "People", and confirm that link pages have been created for each of the people list pages. If you edit one of these special link pages, you should only see a title and a url for the page within the site. These special pages are used to manage the site's navigation menu; the actual display of people on the list page is handled via code. Try changing the title and see if the name of the page in the site's navigation menu also changes.

also in the admin site, check out the menu on the left called "Snippets". this is new; it's where small pieces of text (like page intros) will be stored and edited. you should be able to find the page intro snippet for the "students" and "affiliates" list pages here; go ahead and edit them and see if your changes are reflected when you visit the page. you might need to refresh multiple times for the changes to take effect.

# dev notes
**see PR #269**

see also (and depends on) #186. The new `PersonIndexPage` will probably be an abstract `Page` which we then subclass to pick a particular queryset/filter and create individual pages like "staff", "students", etc.

- [x] `PersonQuerySet` (old `ProfileQuerySet` - now operates at level of `Person`)
- [x] `PersonListPage` (uses QS methods to replace current person views)
- [x] update views for profile -> person
- [x] update templates for profile -> person
[...]

Issue #253, opened 2020-12-08
As a content editor, I want existing people migrated to wagtail so that I can manage them in the new system.


thatbudakguy:
# testing notes
this one is pretty straightforward to test - we just want to make sure that everyone who has a visible profile on the CDH site still has one, and that it looks more more less the same. you can try it out by finding the URLs of current profiles like https://cdh.princeton.edu/people/camey-vansant/, and then comparing the new version (https://cdh-test-web.princeton.edu/pages/people/camey-vansant/). probably a good idea to try at least one of current/past staff, students, faculty, etc.

everything should look the same _except_ for attachments (if people had a pdf of a resume, for example), which will look weird until we finish #246, so ignore those.

note that how people appear on the list pages (like `/people/staff`) isn't covered by this issue.

# dev notes
depends on #186

cvansant:
The menu items that are on the left of some staff pages (e.g. Twitter, ORCID, website) are in a different order on the test site than on the current live site. Here is the [current MM page](https://cdh.princeton.edu/people/meredith-martin/) and the [test site MM page](https://cdh-test-web.princeton.edu/pages/people/meredith-martin/).

I notice two differences between the two Kate Carpenter pages. Here's the [current site](https://cdh.princeton.edu/people/kathryn-carpenter/) and here's the [test site](https://cdh-test-web.princeton.edu/pages/people/kathryn-carpenter/). The photo color looks slightly different. Also, for her blog posts, the test site attributes the posts to "kbc." For Elena Ryan, the profile photos are two different sizes: [current site](https://cdh.princeton.edu/people/elena-m-ryan/) and [test site](https://cdh-test-web.princeton.edu/pages/people/elena-m-ryan/).

Also, I don't know who to test faculty since they just link to outside pages.

thatbudakguy:
Interesting about the link order; I think as part of Kevin's work on #181 we will be able to make the order controllable in the admin so for now the difference doesn't matter.

Re: the Kate image, super weird!! I think this might be related to the fact that some of the images are corrupted and missing EXIF data; that's the only thing I can think of that would cause them to have slight discolorations or changes like that. There's not really anything we can do because the file was modified, but I guess this shows what the result of that is. Ping @rlskoeser if you didn't see it already.

About the blog post author listings, that will be fixed in #185, but good to have a reminder that we'll need to update it here (blog posts aren't yet associated with the new `Person` and are still tied to user accounts, which is why they say `kbc`).

[...]

Issue #186, opened 2019-12-20
As a moderator, I want to create and edit people in wagtail so that I can manage them just like pages.


thatbudakguy:
# testing notes
## as a moderator
to test this, log into the new admin site (`/cms`) and check to see if you have a menu on the left side called "People". 

1. Make a new person by going to the "People" menu, choosing "People", and clicking "new person" at the top right. Feel free to also give the person a picture, and assign them one (or multiple) CDH positions at the bottom.
2. Create a profile for your new person by going to the "People" menu, choosing "Profiles", and clicking "add profile" at the top right. Associate the profile with the person you made in step 1, but feel free to upload a _different_ image, and also give the profile a title that isn't the same as the person's name (add a middle initial, for example).
3. Visit the new profile by clicking "preview" at the bottom (by "publish"). Check that the profile is visually similar to current CDH profiles (e.g. your own profile on the current site). The picture displayed should be the picture you uploaded to the profile (not the person) and the name shown at the top should be the profile's title (not the person's name).

you can also manage the list of available CDH position titles. you should see it as "Titles" underneath "People" on the left.

1. Create a new CDH position title by going to the "People" menu, choosing "Titles", and clicking "add title" at the top right. Don't worry about "sort order" for now since it's only relevant for #267; anything is fine.
2. Give the person that you created in step 1 above (as an editor) this new title.
3. Repeat step 3 above and check that the person's profile shows their new job title.

# dev notes
For `Person`, how we can make it orderable and what type of chooser block will we need (maybe extend)? to add it to other pages? We want to preserve the Django user model, along with some functionality to populate it from LDAP, but no longer make Person inherit it.

For person index pages, see #267.
## models
- [x] `Title` (migrations: change user to person, same as membership)
- [x] `Position` (migrations: change user to person, same as membership)
- [x] `Person` (no longer inherits `User` - see board for fields; optional 1:1 with user)
- [x] `Membership` and `Project` (migration: change user to person)
- [x] `PersonResource` (will become wagtail `PersonRelatedLink` via #181)
- [x] new `ProfilePage` (based on wagtail `Page` - migrate profile-specific fields and link to Person; see board for fields; required 1:1 with person)
- [x] update `init_profile_from_ldap` to create new person 1-1 and populate fields
## admin
- [x] configure `Person` in wagtail admin
- [x] configure `Title` in wagtail admin
- [x] make `Position` inline-editable on `Person` in Wagtail
[...]

Issue #266, opened 2020-12-22
revisit image sizing guidelines


thatbudakguy:
the current sizing strategy has us serving 5000px wide images on desktop/tablet retina screens, which we probably don't need to do. in light of the new strategy we debuted on _Startwords_, maybe we can take another look and revise our strategy for this project.

for reference, see the [startwords `figure` shortcode](https://github.com/Princeton-CDH/startwords/blob/d4d27df52c2150053c35ef65fe1d498c63d142ee/themes/startwords/layouts/shortcodes/figure.html#L26-L35).

PR #268, opened 2021-01-06
feature/person exodus

thatbudakguy:
This PR addresses #186 and #253, but not #267 (the person list pages). It's a big one!

Some things to note:

- Multiple migrations that un-proxy the old `Person` model and make it independent from django's `User`
- Most properties of `Profile` move to the new `Person` (office location, phone, job title, etc.)
- `Profile` becomes a new wagtail `Page` model called `ProfilePage` (but the old `Profile` still exists, so it can be exodized)
- The admin section for this app is redone so that `ProfilePage`s are managed through a special menu (similar to the current admin) and don't clutter the wagtail page tree - check out `wagtail_hooks.py`
- Some fancy Wagtail admin features are added, such as automatic export of people to excel/csv and inline editing of `Title`s on people
- A special `PeopleLandingPage` is created so that correct page hierarchy can be enforced through code
- I adopted a new convention (criticism welcome) of moving model-related tests for wagtail `Page` models to a file called `test_pages.py` for each app, so that `test_models.py` was less crowded and could focus on non-`Page` models.
- I tried to reduce this app's reliance on a single large fixture, which led to rewriting many tests and a lot of repetition of set-up code. This could be refactored later.


cdhweb/people/models.py
```diff
@@ -371,70 +381,115 @@ def current_title(self):
...
+        if not self.person:
+            raise ValidationError("Profile page must be for existing person.")
+
+
+class PeopleLandingPage(LandingPage):
+    """LandingPage subtype for People that holds ProfilePages."""
+    # NOTE this page can't be created in the page editor; it is only ever made
+    # via a script or the console, since there's only one.
+    parent_page_types = []
+    # NOTE the only allowed child page type is a ProfilePage; this is so that
```
rlskoeser, line 439: Ah. I wondered why we needed a subclass of the landing page. Seems reasonable, I guess, since we don't want profile pages being added elsewhere (I think).

cdhweb/people/models.py
```diff
@@ -371,70 +381,115 @@ def current_title(self):
...
+    # ProfilePages made in the admin automatically are created here. Other
+    # special pages, like profile list pages, have to be created and added
```
rlskoeser, line 441: profile list pages will be added by the exodus script, right? not manually? would creating a profile list page manually ever make sense?

cdhweb/people/models.py
```diff
@@ -371,70 +381,115 @@ def current_title(self):
...
-    profile.job_title = str(ldapinfo.title).split('.')[0]
+    # populate Person fields with data we can pull from ldap, if they are empty
+    person, _created = Person.objects.get_or_create(user=user)
 
-    # always update PU status to current
-    profile.pu_status = str(ldapinfo.pustatus)
+    # set first/last name basic on basic whitespace split; can adjust later
+    first_name, *_others, last_name = str(ldapinfo.displayName).split()
+    person.first_name = person.first_name or first_name
+    person.last_name = person.last_name or last_name
```
rlskoeser, line 474: Nice syntax trick, but I don't think we should be splitting displayName here.

We have an attribute map configured in settings that pucas picks up, which sets first name from `givenName` and last name from `sn` (surname). I think the name fields should be copied from user to person instead, if not already set.

cdhweb/people/models.py
```diff
@@ -371,70 +381,115 @@ def current_title(self):
...
+    # set office location ("street" in ldap)
+    if ldapinfo.street and not person.office_location:
+        person.office_location = str(ldapinfo.street)
 
-    # only do if the person has a title set
-    if ldapinfo.title:
-        # job title in ldap is currently stored as
-        # job title, organizational unit
-        job_title = str(ldapinfo.title).split(',')[0]
-        Title.objects.get_or_create(title=job_title)
```
rlskoeser, line 440: Are we dropping this? I think I'm ok with that (it ended up generating a lot of extraneous titles we didn't need), just want to be explicit about it. Maybe a comment about previous behavior?

cdhweb/people/models.py
```diff
@@ -371,70 +381,115 @@ def current_title(self):
...
+    if ldapinfo.ou and not person.department:
+        person.department = str(ldapinfo.ou)
+
+    # set job title (split and use only first portion from ldap)
+    # NOTE titles for cdh staff are managed via Title/Position instead
+    if ldapinfo.title and not person.job_title:
+        person.job_title = str(ldapinfo.title).split(",")[0]
+
+    # always update PU status to current
+    person.pu_status = str(ldapinfo.pustatus)
```
rlskoeser, line 494: nice

cdhweb/people/templates/people/profile_page.html
```diff
@@ -0,0 +1,59 @@
...
+    {# NOTE equiv to contributors in project #}
+    <ul>
+    {% for resource in page.person.personresource_set.all %}
+        <li><a href="{{ resource.url }}">{{ resource.resource_type.name }}</a></li>
+    {% endfor %}
+    </ul>
+</div>
+
+{% if page.image %}
+    {% image page.image fill-790x632 as profile_img %}
```
rlskoeser, line 30: do we already have later tasks about making these responsive in future?

cdhweb/people/urls.py
```diff
@@ -16,5 +16,4 @@
     url(r'^faculty/$', RedirectView.as_view(url='/people/affiliates/', permanent=True)),
     # redirect from /people/postdocs -> /people/staff
     url(r'^postdocs/$', RedirectView.as_view(url='/people/staff/', permanent=True)),
-    url(r'^(?P<slug>[\w-]+)/$', views.ProfileDetailView.as_view(), name='profile'),
```
rlskoeser, line 19: Removed from url config because now handled by wagtail page routing?

Do we need to do anything about that? (Anything to be aware of or note as we move forward with the wagtail exodus?)

cdhweb/people/wagtail_hooks.py
```diff
@@ -0,0 +1,47 @@
...
+class PersonAdmin(ThumbnailMixin, ModelAdmin):
+    model = Person
+    menu_icon = "group"
+    list_display = ("first_name", "last_name", "admin_thumb", "current_title",
+                    "cdh_staff")
+    search_fields = ("first_name", "last_name", "user__username")
+    list_filter = ("pu_status", "cdh_staff")
+    list_export = ("first_name", "last_name", "admin_thumb", "current_title",
+                   "cdh_staff")
+    export_filename = "cdhweb-people"
```
rlskoeser, line 16: cool

cdhweb/people/wagtail_hooks.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class TitleAdmin(ModelAdmin):
+    model = Title
+    menu_icon = "order"
+    list_display = ("title", "sort_order", "num_people")
+    search_fields = ("title",)
+
+# remove ProfilePages from the regular wagtail page editor and create a special
+# admin area for them underneath the "people" section
```
rlskoeser, line 27: sounds like a good solution

cdhweb/resources/models.py
```diff
@@ -39,20 +35,19 @@ def excerpt(self):
...
-
-class UserResource(models.Model):
-    '''Through-model for associating users with resource types and
+class PersonResource(models.Model):
+    '''Through-model for associating people with resource types and
     corresponding URLs for the specified resource type.'''
     resource_type = models.ForeignKey(ResourceType, on_delete=models.CASCADE)
-    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    person = models.ForeignKey("people.Person", on_delete=models.CASCADE)
     url = models.URLField()
```
rlskoeser, line 43: maybe this should be in the people apps instead of resources

cdhweb/resources/models.py
```diff
@@ -82,8 +78,9 @@ def __str__(self):
...
     tagline = models.CharField(max_length=255)
     image = FileField(verbose_name="Image",
-        upload_to=upload_to("resources.landing_pages.image", "resources"),
-        format="Image", max_length=255, null=True, blank=True)
+                      upload_to=upload_to(
+                          "resources.landing_pages.image", "resources"),
+                      format="Image", max_length=255, null=True, blank=True)
 
 
 class PublishedQuerySetMixin(models.QuerySet):
```
rlskoeser, line 86: This will go away, right? Should we add a deprecated warning or will it naturally be removed once we exodize all the relevant pieces?

cdhweb/settings.py
```diff
@@ -24,7 +24,8 @@
                 (_("Media Library"), "media-library"),
                 "resources.Attachment")),
     ("Events", ('events.Event', 'events.Location', 'events.EventType')),
-    ('Profiles', ('people.Person', 'people.Position', 'people.Title')),
+    ('Profiles', ('people.Position', 'people.Title')),
+    # ('Profiles', ('people.Person', 'people.Position', 'people.Title')),   #
```
rlskoeser, line 28: still needed?

cdhweb/settings.py
```diff
@@ -412,6 +414,7 @@
     # local method to populate profile fields based on available
     # ldap information
     'EXTRA_USER_INIT': 'cdhweb.people.models.init_profile_from_ldap'
+    # TODO: update for new person model
```
rlskoeser, line 417: can remove this todo

cdhweb/pages/management/commands/exodus.py
```diff
@@ -238,6 +285,32 @@ def form_pages(self):
         'root': Collection.objects.get(pk=get_root_collection_id())
     }
 
+    def profile_pages(self):
+        """Exodize all existing Profiles to new ProfilePage model."""
+        # all the fields on Profile that moved to Person will have been handled
+        # by django migrations; we just need to create new Page models
+        people_landing = PeopleLandingPage.objects.first()
+        if not people_landing:
```
rlskoeser, line 293: should this warn or error? it shouldn't happen, since we create it first, but seems like we would want to know if it didn't run

cdhweb/pages/management/commands/exodus.py
```diff
@@ -238,6 +285,32 @@ def form_pages(self):
...
+            return
+        for profile in Profile.objects.filter(user__person__isnull=False):
+            # NOTE not sure why Profile.user.person fails; works in console?
```
rlskoeser, line 296: still an issue?

cdhweb/pages/management/commands/exodus.py
```diff
@@ -238,6 +285,32 @@ def form_pages(self):
...
+            profile_page = ProfilePage(
+                person=person,
+                title=profile.title,
+                image=self.get_wagtail_image(profile.image) if profile.image else None,
+                education=profile.education,
+                bio=json.dumps([
+                    {"type": "migrated", "value": profile.bio},
+                ])
+            )
+            # added as child of people landing page so slugs are correct
```
rlskoeser, line 307: Where does slug get set?

cdhweb/pages/management/commands/exodus.py
```diff
@@ -262,12 +335,16 @@ def image_exodus(self):
         # delete all images and collections prior to run
         # (clear out past migration attempts)
         Image.objects.all().delete()
-        Collection.objects.exclude(pk=get_root_collection_id()).delete()
+        # Collection.objects.exclude(pk=get_root_collection_id()).delete()
+        # fix collection tree so root collection is up to date
+        # FIXME: this is still not working; seems to error on alternate runs
+        # Collection.fix_tree()
```
rlskoeser, line 341: Still an issue? Do we want to do anything about this?

If it is still a problem, I'm inclined to not delete collections. They shouldn't change once the logic is correct.

cdhweb/pages/management/commands/exodus.py
```diff
@@ -353,5 +430,17 @@ def get_media_files(self):
     def get_wagtail_image(self, image):
         # get the migrated wagtail image for a foreign-key image
         # using image file basename, which is migrated as image title
-        return Image.objects.get(title=os.path.basename(image.name))
-
+        try:
+            return Image.objects.get(title=os.path.basename(image.name))
+        except Image.DoesNotExist as err:
+            print("Error locating image %s: %s" % (image, err))
+            return None
```
rlskoeser, line 437: Do we need to do some error handling related to this, and is that work tracked anywhere? (Are these some of the images that aren't migrating properly?)

cdhweb/pages/management/commands/exodus.py
```diff
@@ -353,5 +430,17 @@ def get_media_files(self):
...
+
+    def person_image_exodus(self):
+        # for people with profile, set larger image as wagtail image for person
+        for profile in Profile.objects.filter(user__person__isnull=False):
+            # NOTE not sure why Profile.user.person fails; works in console?
+            person = Person.objects.get(user=profile.user)
+            if profile.image:
```
rlskoeser, line 444: Could maybe filter on image in the queryset? (Although I think I had problems with a similar filter somewhere, not sure if it's worth it)

Do we have any cases where someone has a thumbnail but no larger image, maybe for people who are only on list views? Should we use the thumbnail as a fallback if image is not set?

cdhweb/pages/models.py
```diff
@@ -157,6 +155,13 @@ def get_context(self, request):
         '''Add featured updates, projects, and events to the page context.'''
         context = super().get_context(request)
 
+        # FIXME an ImportError is thrown when classes in this module are
```
rlskoeser, line 158: weird. I can take a look

rlskoeser (commented):
Ok. I _think_ I have looked at all the files (what a monster of a PR).  Haven't run anything locally yet, there may be a few more things I need to look at when I do.

Happy with your decision to split out unit tests more, they needed to be broken out more.

cdhweb/people/migrations/0011_populate_nonproxy_person.py
```diff
@@ -0,0 +1,56 @@
...
+            person_info.update({
+                'cdh_staff': profile.is_staff,
+                'job_title': profile.job_title,
+                'department': profile.department,
+                'institution': profile.institution,
+                'pu_status': user.profile.pu_status
+            })
+        Person.objects.create(**person_info)
+
+    # NOTE: prune users who don't have login access? later?
```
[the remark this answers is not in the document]
thatbudakguy: I was looking at the google doc tracking the migration progress and we do have a note to do this post-3.0. I can make a chore too

cdhweb/people/migrations/0011_populate_nonproxy_person.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # TODO: do we care about log entries for person creation?
+    # get script user and log entry model to document changes
+    # script_user = User.objects.get(username=settings.SCRIPT_USERNAME)
+    # LogEntry = apps.get_model('admin', 'LogEntry')
+    # ContentType = apps.get_model('contenttypes', 'ContentType')
```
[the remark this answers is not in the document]
thatbudakguy: I think we sectioned off all the log entry creation stuff into #247, although looking at it, that issue only mentions Pages. maybe we should expand it?

cdhweb/people/migrations/0015_person_cluster_titles.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.2.17 on 2020-12-23 16:50
+
+from django.db import migrations
+import django.db.models.deletion
+import modelcluster.fields
```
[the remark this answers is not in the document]
thatbudakguy: Nope, it's just a dependency of wagtail. It gets used to implement wagtail's inline-editing, so any place you see `ParentalKey` on a model or `ClusterableModel` that's what they're from.

cdhweb/people/migrations/0016_person_phone_office.py
```diff
@@ -0,0 +1,23 @@
...
+        migrations.AddField(
+            model_name='person',
+            name='office_location',
+            field=models.CharField(blank=True, help_text='Office location (optional)', max_length=255),
+        ),
+        migrations.AddField(
+            model_name='person',
+            name='phone_number',
+            field=models.CharField(blank=True, help_text='Office phone number (optional)', max_length=50),
+        ),
```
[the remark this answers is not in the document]
thatbudakguy: I figured we wanted to just move as much as possible to person, but probably not hard to undo this change. It would just be moving the model fields, moving these operations to the `ProfilePage` migration, and then updating the exodus script accordingly.

cdhweb/people/models.py
```diff
@@ -39,101 +47,100 @@ def num_people(self):
     num_people.short_description = '# People'
 
 
-class Person(User):
-    # NOTE: using a proxy model for User so we can customize the
-    # admin interface in one place without having to extend the django
-    # default user model.
+class Person(ClusterableModel):
+    # in cdhweb 2.x this was a proxy model for User;
+    # in 3.x it is a distinct model with 1-1 optional relationship to User
```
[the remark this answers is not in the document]
thatbudakguy: it's required for inline-editing in wagtail. for a detailed example of another setup that uses `ClusterableModel` and `ParentalKey` check out the explanation I wrote on #181 for kevin.

cdhweb/people/models.py
```diff
@@ -371,70 +381,115 @@ def current_title(self):
...
+        """Validate that a Person was specified for this profile."""
+        # NOTE we can't cascade deletes to wagtail pages without corrupting the
+        # page tree. Instead, we use SET_NULL, and then add a delete handler
+        # to Person to delete the corresponding profile manually. We still need
+        # to make sure that it's impossible to create a ProfilePage without a
+        # corresponding Person. More info:
+        # https://github.com/wagtail/wagtail/issues/1602
+        if not self.person:
+            raise ValidationError("Profile page must be for existing person.")
+
```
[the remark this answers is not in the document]
thatbudakguy: The pull request only addresses usage of `models.PROTECT`, afaik - wagtail has no plans to address this for `CASCADE`. I saw someone mentioned that it would be resolved if they stopped using `django-treebeard` for maintaining the page tree, but that seems like a huge change...my guess is it'll be here for awhile. I didn't see any problems in practice, but in fairness I didn't try `CASCADE` because I saw the warning and figured why mess around, since we've already managed to corrupt the page tree doing other things 😬

cdhweb/people/models.py
thatbudakguy: sorry, that's what I meant by "manually"! not very clear

cdhweb/people/models.py
thatbudakguy: Yeah, it seemed like it would cause confusion because `Title`s are largely for positions at CDH, and this would basically create a new one for every possible job title at Princeton (which would explain why there are so many weird ones in the current system). I think setting `job_title` on the `Person` is plenty. I'll add a comment

cdhweb/people/templates/people/profile_page.html
thatbudakguy: imo would be covered by #266

cdhweb/people/urls.py
thatbudakguy: correct. eventually i think the only thing in these urlconfs will be sitemaps, using wagtail's contrib.sitemap module. the redirects we can convert into ones stored in the database, so that they can be changed/managed in the admin.

cdhweb/resources/models.py
thatbudakguy: yep - it got moved in #269

cdhweb/resources/models.py
thatbudakguy: We could add a deprecation warning, but I think it will naturally stop being used once we exodize all the relevant pieces, at which point we can just delete. The wagtail equivalent is just `.live` (on a page model) or `.live()` (on a queryset of pages).

cdhweb/settings.py
thatbudakguy: nope!

cdhweb/pages/management/commands/exodus.py
thatbudakguy: the reason this check is here is for when the database has a fixture loaded but is otherwise empty, like when you need to exodize an old fixture so that it can be used with wagtail. it's legit for there to not be a people landing page in such cases, but we do want to proceed and do everything else we can

cdhweb/pages/management/commands/exodus.py
thatbudakguy: i think so, yeah. probably can remove the comment since it works as-is, I just didn't understand why `Profile.user.person` didn't work

cdhweb/pages/management/commands/exodus.py
thatbudakguy: hmm. right now it would be getting generated automatically from the title, but we don't want that! I need to add a call to `convert_slug()` to use the old one, nice catch

cdhweb/pages/management/commands/exodus.py
thatbudakguy: Still a problem and agreed. Let's just not worry about deleting them

cdhweb/pages/management/commands/exodus.py
thatbudakguy: I'm not sure...I don't understand the reasons why images would be missing well enough; I'd defer to you (but yes that's what these are). Overall it would be good to have more real logging/handling for the exodus script (maybe we could improve the messaging about the corrupt EXIF data too).

cdhweb/pages/management/commands/exodus.py
thatbudakguy: I was running into cases where people didn't have an image when I expected them to; maybe that's why...I'm not sure, but I bet that does happen. Probably we should use the thumbnail as a fallback, yep

cdhweb/pages/models.py
thatbudakguy: i think this is resolved in #269 fwiw; i don't remember what the solution was...maybe it was because some things got moved between modules or imports changed

cdhweb/people/models.py
thatbudakguy: how would i access `first_name` and `last_name` in this context? i see the settings but not sure how it works.

re: copying from user to person, wouldn't this already happen as a result of the migrations? i don't think that user has a `first_name` or `last_name` after that, because it's just a regular django user model .

cdhweb/people/migrations/0011_populate_nonproxy_person.py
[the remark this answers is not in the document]
rlskoeser: sounds good

cdhweb/people/migrations/0011_populate_nonproxy_person.py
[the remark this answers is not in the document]
rlskoeser: IIRC #247 is about transferring existing log entries when we exodize records; this todo is about (potentially) creating log entries to document that these person entries were created via script, which I think needs to be done here (if we do it)

cdhweb/pages/management/commands/exodus.py
rlskoeser: Yes! We need to make sure we use the same slugs that we have now

cdhweb/pages/management/commands/exodus.py
rlskoeser: ok, good point — probably the real place to look is the image exodus portion. Should that be handled during tested or should we add a chore to investigate the handful of images that are not being converted?

cdhweb/people/migrations/0011_populate_nonproxy_person.py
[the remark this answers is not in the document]
thatbudakguy: gotcha. that seems simple enough, I'll give it a shot - agree we should do it

cdhweb/pages/models.py
rlskoeser: that makes sense! will be good to have it resolved

cdhweb/people/migrations/0011_populate_nonproxy_person.py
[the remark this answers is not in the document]
rlskoeser: Great. I think I have examples elsewhere if it would be helpful

cdhweb/people/models.py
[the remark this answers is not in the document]
rlskoeser: ok

cdhweb/people/models.py
rlskoeser: Copying from user to person only happens in migrations for existing records; this method is intended primarily for populating new records.

The user model has existing `first_name` and `last_name` properties that are part of the regular django user model (see https://docs.djangoproject.com/en/3.1/ref/contrib/auth/#fields), which will already be populated from LDAP. It makes more sense to me to use those fields rather than getting them directly from ldapinfo, since if the fields we need change we want to be able to change in just one place.

cdhweb/pages/management/commands/exodus.py
thatbudakguy: probably chore, I don't think it should block release of the image exodus feature.

cdhweb/people/templates/people/profile_page.html
rlskoeser: ok, fair. I added a preliminary check list to that issue so we'll remember it is included

cdhweb/people/migrations/0016_person_phone_office.py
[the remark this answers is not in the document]
rlskoeser: I think we want the person / profile split to match what is displayed on person list views and profile detail pages. (I think this was a big part of the confusion before, the way they were split didn't match that).

cdhweb/people/models.py
thatbudakguy: didn't realize that django's user model had these fields too! makes sense to just copy them on `Person` init then. does it matter that we would have:
- `User.first_name` / `User.last_name`
- `Person.first_name` / `Person.last_name`
- `Profile.title`

which can all be different? I guess nobody's really going to change the ones on `User`.

cdhweb/people/migrations/0016_person_phone_office.py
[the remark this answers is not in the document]
thatbudakguy: That makes sense; I think we want it as intuitive as possible

cdhweb/pages/management/commands/exodus.py
rlskoeser: That makes sense to me (as long as we make it part of 3.0); and we may figure more things out once we start doing acceptance testing and getting another set of eyes on all of this.

cdhweb/people/models.py
rlskoeser: Since `Person.user` is optional, we need to duplicate the fields. I think the names on `user` are mostly for convenience, so we know which user accounts we're managing. It might be nice to sync them from `person` to `user` on save, but not sure it's worth the hassle.

Is `title` required on profile page? The title field was required for mezzanine, and I think it was sometimes a point of confusion where you needed to edit that. It would be nice if we could get rid of that one, so it wouldn't be possible for the profile page title to be different from the person's name. (I wonder if we have any cases where they differ now?)

thatbudakguy: It is required for `Page` subtypes. I actually implemented exactly that, until I realized that it differs for a decent number of people - e.g. grant, who is `Grant R. Wythoff` in his profile. So unfortunately I don't think we can do it unless we want to take away those middle initials.

rlskoeser: I noticed the middle initial on his a while back and wondered if he wanted it there but never asked. I _think_ this is based on a difference between the first and last names and display name in LDAP. But we might still want the flexibility to have a full display name that _can_ be different from first and last...

cdhweb/people/migrations/0011_populate_nonproxy_person.py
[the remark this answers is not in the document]
thatbudakguy: created #270

cdhweb/pages/management/commands/exodus.py
thatbudakguy: created #271

cdhweb/people/migrations/0011_populate_nonproxy_person.py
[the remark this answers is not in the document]
thatbudakguy: Added log entry creation, but I noticed that there's no way to view them in the wagtail admin (that I can find). Not sure how important this is though since we can do it in console. I modified `admin.py` so it would work enough to go in and verify that they are indeed there.

cdhweb/people/migrations/0011_populate_nonproxy_person.py
```diff
@@ -28,15 +28,22 @@ def create_nonproxy_persons(apps, schema_editor):
...
-    # LogEntry = apps.get_model('admin', 'LogEntry')
-    # ContentType = apps.get_model('contenttypes', 'ContentType')
+        person = Person.objects.create(**person_info)
+
+        # log that the person was created via this migration script
+        LogEntry = apps.get_model('admin', 'LogEntry')
+        ContentType = apps.get_model('contenttypes', 'ContentType')
+        script_user = User.objects.get(username=settings.SCRIPT_USERNAME)
+        person_contenttype = ContentType.objects.get(
+            app_label='people', model='person')
```
rlskoeser, line 38: These lines should be moved outside the for loop, but otherwise this looks good.

rlskoeser (commented):
Updates since my last review look good; just one comment on some lines that I think are inside a for loop and shouldn't be.

cdhweb/pages/management/commands/exodus.py
thatbudakguy: I counted 57 cases of this, actually, with:
```py
Profile.objects.filter(image="").exclude(thumb="").count()
```
added a check to use thumbnails as a fallback.

cdhweb/people/migrations/0011_populate_nonproxy_person.py
thatbudakguy: done

cdhweb/pages/management/commands/exodus.py
thatbudakguy: added a call to `convert_slug()`

kmcelwee (approved):

PR #272, opened 2021-01-08
move office loc and phone number back to profilepage

thatbudakguy:


rlskoeser (commented):
Sorry you did this work before we discussed it! After considering it more, I think the benefit of having the fields auto-populated from ldap outweighs the value of having them only on the profile page where they're actually used.  (I apologize for sending you down a rabbit trail) Setting them manually seems not worth the effort (and is the kind of thing that would be easily skipped if we rely on it).

I think we should postpone any decision on removing phone numbers and office numbers from the profile page; that's easy enough to do later, but it would be nice to consider properly both in terms of functionality and visual display.

PR #275, opened 2021-01-13
Port CaptionedImageBlock and SVGImageBlock from s&co

thatbudakguy:


rlskoeser (approved):
Changes look good to me.

How much additional effort to include the half-width left/right styles from PPA ?

thatbudakguy:
Definitely more since it involves CSS we don't have on cdhweb, I think (and thus is way less likely to "just work"). I stayed away from `LinkableSectionBlock` for the same reason, but could be good for v3.1.

rlskoeser:
> Definitely more since it involves CSS we don't have on cdhweb, I think (and thus is way less likely to "just work"). I stayed away from `LinkableSectionBlock` for the same reason, but could be good for v3.1.

I thought maybe the css would be easily portable from PPA (I thought it was custom). But agree, 3.1 makes sense. (And maybe then we can extend the ppa styles further for quarter-width images like were added to some pages manually)

PR #276, opened 2021-01-14
Exodize person list views

rlskoeser:
- refactor old `ProfileQuerySet` to new `PersonQuerySet` (existing work by @thatbudakguy pulled in from other person list branch)
- updates existing profile list views to person lists views that work with the new person model
- shift person card display logic from template to view
- new unit tests using pytest fixtures to create various person records
- adds wagtailmenus and creates a `LinkPage`
- new `PageIntro` snippet with foreign key to `LinkPage`
- new context processor to add `PageIntro` to context if there is one for the current request
- update exodus script to generate LinkPages and optionally PageIntro for the mezzanine pages corresponding to our person list views

I didn't write tests for the speaker list view yet because I think it might be easier to wait on event exodus/refactor (events are still linked to user instead of person).


cdhweb/pages/context_processors.py
```diff
@@ -0,0 +1,13 @@
...
+def page_intro(request):
+    '''Template context processor: if there is a PageIntro snippet
+    for this page, add it to the context for display.'''
+    # wagtail stores link url without leading and trailing slashes,
+    # but requests to django view urls include them; strip them off to match
+    page_intro = PageIntro.objects.filter(
+        page__link_url=request.path.strip('/')).first()
+    if page_intro:
+        return {'page_intro': page_intro}
+    return {}
```
thatbudakguy, line 13: nice, this is cool!

cdhweb/people/tests/conftest.py
```diff
@@ -0,0 +1,108 @@
+from datetime import date, timedelta
```
thatbudakguy, line 1: can we just call this file "fixtures"?

cdhweb/people/tests/test_models.py
```diff
@@ -85,6 +86,280 @@ def test_str(self):
         assert str(self.person) == "tom jones"
 
 
+class TestPersonQuerySet(TestCase):
+
+    def setUp(self):
+        """create testing data"""
+        # FIXME why is this necessary?
+        Title.objects.all().delete()
```
thatbudakguy, line 94: curious if you ran into this also, it seemed liked some titles were already in the database when tests ran for some reason

thatbudakguy (approved):
nice work wrangling all these complicated views!

cdhweb/people/tests/conftest.py
rlskoeser: no, conftest is a special pytest name that gets loaded automatically

cdhweb/people/tests/test_models.py
rlskoeser: I don't remember this, must be from your code that I cherry-picked over. Do we preload titles anywhere in data migrations? maybe for exec?

cdhweb/people/tests/conftest.py
thatbudakguy: oh! good to know

cdhweb/people/tests/test_models.py
thatbudakguy: yep, that was it! we create them in `cdhweb/people/migrations/0006_exec_committee_role.py`. good catch. I guess we should just switch the tests/fixtures to `get_or_create()`.

PR #277, opened 2021-01-14
Feature/project exodus

thatbudakguy:
- rename `Project` and models with FKs to it to `Old[Model]`
- create a migration
- delete some things (like old `admin.py`s) that were throwing errors as a result

ignore the commit history, i messed some things up.


cdhweb/projects/migrations/0017_rename_mezz_models.py
```diff
@@ -0,0 +1,99 @@
...
+    dependencies = [
+        ('sites', '0002_alter_domain_unique'),
+        ('people', '0021_peoplelandingpage'),
+        ('resources', '0008_userresource_to_personresource'),
+        ('taggit', '0003_taggeditem_add_unique_index'),
+        ('projects', '0016_project_member_to_person'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
```
rlskoeser, line 22: If we proceed with this approach, we need to make sure this is a rename and not a create!

Issue #249, opened 2020-12-08
As a moderator, I want existing related links migrated to wagtail so that I can manage them in the new system.


thatbudakguy:
# testing notes
this is a companion to #181; you may find it easier to test that first. see that issue's notes for a description of what related links are and how they work.
## as a user
- visit a few project pages on the test site and check that all their related links are displayed, and that the display matches the project page on the current production site. try a sponsored project, a staff project, and a working group.
- visit a few profiles on the test site and check that all their related links are displayed, and that the display matches the profile on the current production site. staff and students are generally the only ones with profiles.
- find a person who doesn't have a profile (easiest with faculty affiliates) but who does have a website link. check that clicking them in the list view on the test site takes you to their website, just like it does on the production site.
# dev notes

kmcelwee:
Testing Notes:
- [ ] Find a project with a related link on prod ([like the Natasha's NEH grant](https://cdh.princeton.edu/projects/new-languages-nlp-building-linguistic-diversity-digital-humanities/)) Make sure that that also exists in QA
- [ ] Compare related links on profiles for prod and QA and ensure that they are the same.

cvansant:
@thatbudakguy The links aren't visible as far as I can tell. For instance, on the current site, Zoe has links to Github, Twitter, and her website: https://cdh.princeton.edu/people/zoe-g-leblanc/

[Here ](https://test-web.cdh.princeton.edu/pages/people/zoe-g-leblanc/)there are no such links

manawinters:
@thatbudakguy , the only projects that appear in the test site are "Past Projects" under "Sponsored Projects." [Staff Projects](https://test-web.cdh.princeton.edu/projects/staff/) is blank and [DH Working Groups](https://test-web.cdh.princeton.edu/projects/working-groups/) goes to the description.

manawinters:
Links and display of staff and students looks good. The only difference between the two sites are the Recent Posts - in the [test site](https://test-web.cdh.princeton.edu/pages/people/rebecca-sutton-koeser/) there is the post title, user name (rkoeser), date, and the first sentence of the post. In the [production site](https://cdh.princeton.edu/people/rebecca-sutton-koeser/), it just shows the post title and date.

There are no Affiliates listed on the [test site](https://test-web.cdh.princeton.edu/people/affiliates/).

Issue #181, opened 2019-12-19
As a moderator, I want to create and edit related links in wagtail so that I can associate them with projects and people.


thatbudakguy:
# testing notes
fyi - "related links" were previously called "resources"; we've renamed them to be a bit more clear. they allow you to add a particular link to a person or project, which will show up on the corresponding page or profile.

the "website" related link is special - if you add a website to a person, clicking on them on a list page will take you to that website (unless they already have a profile on the CDH site). we do this often for faculty, who have profile pages on departmental websites that we want to link to instead of making them a profile on our site. if you add a website to a project, the website will be displayed in a special way on the project's page, showing its URL.

### projects
1. go to the admin site and edit any project. scroll down to the bottom until you find the "related links" section.
2. add a new related link of any type. if the project doesn't have a website link, add a website link too.
3. visit the project's page by clicking "preview". you should see the link you added, as well as the website link, and the website link should show the URL you entered.
### people
1. go to the admin site and edit any non-staff CDH person. scroll down to the bottom until you find the "related links" section.
2. add a website link for this person.
3. find the person on whichever person list page they will show up (e.g. students) and click them - you should go to the website URL you added.
4. go back and edit any CDH staff member, repeating step 1 above.
5. add any kind of related link for this person you like, including a website link.
6. visit the person's profile by clicking them on the staff list page. this should take you to their profile, _not_ the website URL you added (we only use that URL if you don't have a profile on our site).
7. you should see any other related links you added on the person's profile, shown on the left.

# dev notes
This is ready to be worked on once #268 is merged.

## background
**n.b.**
You will see the term "Resource" used throughout the codebase, but the title of this issue uses the term "Related Link"; this is the term we want to use going forward because it's more descriptive. Thus related links for people will be renamed from `PersonResource` to `PersonRelatedLink`, etc.

We currently allow associating arbitrary URLs (related links) with two models in cdh-web: `Person` and `Project`. This is used, for example, to display a list of relevant links on a person's profile page:

<img width="554" alt="Screen Shot 2021-01-08 at 11 08 22 AM" src="https://user-images.githubusercontent.com/4924494/104036860-efd17c00-51a1-11eb-93b1-886793eefde0.png">

[...]

Issue #182, opened 2019-12-19
As a moderator, I want to create and edit projects in wagtail so that I can manage projects just like other pages.


thatbudakguy:
# testing notes
## as a moderator
to test this, log into the new admin site (`/cms`) and check to see if you have a menu on the left side called "Projects".

1. Make a new project by going to the "Projects" menu, choosing "Project", and clicking "new project" at the top right. Feel free to also give the project an image/thumbnail and create some fake grants for it at the bottom. Make sure to add at least one membership to the project as well, down by where the grants are.
2. Visit the project's page by clicking "preview" at the bottom. Check that the project page is visually similar to current CDH project pages. The picture displayed should be the main image you uploaded to the project (not the thumbnail). The page should display the membership you added to the project and any grants you added.
3. Go back to the projects menu and choose "memberships". Are you able to find the membership you created when you made your new project in step 1? Try adding more memberships to your new project using this menu. Going forward, you can add memberships here, or directly on the project - whichever is easier.
4. Visit the project page again and check that the memberships you added in step 3 are now visible also.

moderators can also manage the list of available project roles, and manage the list of available grant types.

1. Create a new project role by going to the "Project" menu, choosing "Roles", and clicking "add role" at the top right. Don't worry about "sort order"; anything is fine.
2. Give any of the project members that you created in the steps above this new role.
3. Check the project page you created above and see if the list of members has been updated with the new role.
4. Create a new grant type by going to the "Project" menu, choosing "Grant Types", and clicking "add grant type".
5. Add a new grant to the project you created above like you did in step 1 above.
6. Check the project page again and see if your new grant is listed.

# dev notes
<s>going to leave `Membership` and our project association structure largely unchanged for now, but will revisit (potentially in v3.1) to improve it - likely by associating `Person`s to `Project`s directly.</s> done
## models
- [x] `Project` (becomes `Page` subtype, uses wagtail image fields, adds tags)
- [x] `ProjectQuerySet` (minimal changes)
- [x] `GrantType` (minimal changes)
- [x] `Grant` (minimal changes)
- [x] `Role` (minimal changes)
- [x] `Membership` (minimal changes)
- [x]  `ProjectLandingPage` - create special `LandingPage` subtype; cf. `PeopleLandingPage`
## admin
- [x] create a new section in the admin for Project-related models using `ModelAdmin`
- [x] set up the main admin for `Project`s (use `ProfilePage` as an example)
[...]

PR #278, opened 2021-01-15
Feature/181 related link

kmcelwee:
This PR addresses both #181 and #249

* Move ResourceType from resource to pages app and rename to RelatedLink 1ec8176
* Add RelatedLink model 76b4eb3
* Rename ProjectResource to ProjectRelatedLink 67c774e
* Have ProjectRelatedLink inherit from RelatedLink and migrate ae59c51
* Move PersonResource to people app and rename to PersonRelatedLink e4fc0ce
* Have PersonRelatedLink inherit from RelatedLink and migrate 8183384
* Add InlinePanels e4502cf

This takes advantage of [SeparateDatabaseAndState](https://docs.djangoproject.com/en/3.1/ref/migration-operations/#django.db.migrations.operations.SeparateDatabaseAndState), which allows us to give database commands without telling django about them. This is necessary to move tables from the `resources` app to `pages` and `people` apps


cdhweb/pages/tests/test_pages.py
```diff
@@ -71,3 +71,9 @@ def test_parent_pages(self):
     def test_subpages(self):
         # only allowed child is content page
         self.assertAllowedSubpageTypes(ContentPage, [ContentPage])
+
+
+class TestRelatedLinkType(WagtailPageTests):
+    def test_str(self):
+        restype = RelatedLinkType(name='twitter')
+        assert str(restype) == restype.name
```
thatbudakguy, line 79: I think this can just be a `SimpleTestCase` rather than `WagtailPageTests` since it's not using any of that functionality. might want to also change the variable name "restype" since i think that's a holdover when when it was called "resource type", but no biggie

cdhweb/people/admin.py
```diff
@@ -2,7 +2,7 @@
 from mezzanine.core.admin import DisplayableAdmin
```
thatbudakguy, line 2: this entire file can be deleted since we're no longer managing the people admin through mezzanine/django. take a look at `wagtail_hooks.py` in the people app and there should be a TODO there about adding inline editing for relatedlinks, and have a look at the docs for [wagtail's `ModelAdmin`](https://docs.wagtail.io/en/stable/reference/contrib/modeladmin/index.html).

cdhweb/people/models.py
```diff
@@ -495,3 +498,9 @@ def init_person_from_ldap(user, ldapinfo):
     # always update PU status to current
     person.pu_status = str(ldapinfo.pustatus)
     person.save()
+
+
+class PersonRelatedLink(Orderable, RelatedLink):
+    '''Through-model for associating people with resource types and
+    corresponding URLs for the specified resource type.'''
+    person = ParentalKey(Person, on_delete=models.CASCADE, related_name="related_links")
```
thatbudakguy, line 506: I've belatedly realized that we don't want `Orderable` here (although having it does nothing). That mixin adds a `sort_order` field, but we already define that field manually on `RelatedLinkType`, and we want it to be editable in the admin so that links are always ordered based on their type (rather than reorderable on each page). So, you can just remove the references to `Orderable` for `RelatedLink` subtypes (and theoretically nothing should happen).

cdhweb/people/templates/people/profile_page.html
```diff
@@ -20,8 +20,8 @@ <h1>{{ page.title }}</h1>
 <div class="links">
     {# NOTE equiv to contributors in project #}
     <ul>
-    {% for resource in page.person.personresource_set.all %}
-        <li><a href="{{ resource.url }}">{{ resource.resource_type.name }}</a></li>
+    {% for resource in page.person.personrelatedlink_set.all %}
+        <li><a href="{{ resource.url }}">{{ resource.type.name }}</a></li>
```
thatbudakguy, line 24: because you set a `related_name` for this field, you should just be able to do `person.related_links.all`, which is a little nicer

thatbudakguy: I'd also change it to `{% for link in ...%}`, `link.url`, etc

cdhweb/projects/admin.py
```diff
@@ -3,11 +3,11 @@
 from mezzanine.core.admin import DisplayableAdmin
```
thatbudakguy, line 3: same deal with this file as the other `admin.py`, it can just be deleted. I'll set up the admin management for `ProjectRelatedLink`s as part of #182.

cdhweb/projects/templates/projects/project_detail.html
```diff
@@ -12,8 +12,8 @@ <h1>{{ project.title }}</h1>
 
 <div class="links">
   <ul>
-    {% for link in project.projectresource_set.all %}
-        {% with link_type=link.resource_type.name %}
+    {% for link in project.projectrelatedlink_set.all %}
```
thatbudakguy, line 15: same deal here, you can use the related name `project.related_links`

cdhweb/resources/admin.py
```diff
@@ -4,7 +4,8 @@
 from mezzanine.pages.models import Page, RichTextPage
```
thatbudakguy, line 4: this file can get deleted too; its admin will all move to `pages`.

thatbudakguy (changes requested):
migration stuff (the hard part) all looks good, just a few comments re: naming things and cleaning up unneeded files.

cdhweb/pages/tests/test_pages.py
rlskoeser: For pytest, it actually doesn't need to extend anything if we're not using any functionality from the test case classes

cdhweb/people/templates/people/profile_page.html
kmcelwee: addressed by 9b415a9

cdhweb/projects/templates/projects/project_detail.html
kmcelwee: addressed by 9b415a9

cdhweb/resources/admin.py
kmcelwee: addressed by 4a10cd49b355fd2fe6aabe0b62b7d63058af6625

cdhweb/projects/admin.py
kmcelwee: addressed by 4a10cd4

cdhweb/people/admin.py
kmcelwee: addressed by 4a10cd4

cdhweb/pages/tests/test_pages.py
kmcelwee: addressed by 6c2e39d0e679047e2728f65cb627db96b65e0405

PR #281, opened 2021-02-03
Person list refinements after first round of testing

rlskoeser:
This branch includes the following fixes related to person lists after first round acceptance testing:
- fix project director label for student affiliates
-  display multiple roles for student affiliates
-  check student page thumbnail display and adjust image resizing logic
- investigate including page intro modification time in last modified /conditional get (determined not to worry about it)
- remove speaker page

I checked the faculty/affiliate link page url and blurb text, this was a data problem in production and I corrected it. (I think the mezzanine page never got moved when the view was moved).

I investigated the student page thumbnail display and decided to enable image focal point detection. It looks like once this is turned on and the libraries are present, wagtail will automatically calculate focal point when images are created — which means its taken care of for us when we generate the images in our exodus script. I added the ubuntu package to our deploy but we will need to test in qa; it's possible we'll have some troubleshooting to do.

I struggled to find a good image crop/focus configuration that worked well for all the images we currently have. There are a couple of images that still look a little strange, might be worth tweaking it a little. In some cases we'll probably want to adjust the focal point manually to get the images to crop a little tighter.

### DIFF
--- a/cdhweb/people/tests/test_views.py
+++ b/cdhweb/people/tests/test_views.py
@@ -7,7 +7,7 @@ from pytest_django.asserts import assertContains, assertNotContains
 
 from cdhweb.people.models import Person
 from cdhweb.people.views import AffiliateListView, ExecListView, \
-    SpeakerListView, StaffListView, StudentListView
+    StaffListView, StudentListView
 
 # NOTE: person factory fixtures in conftest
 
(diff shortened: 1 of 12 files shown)


cdhweb/people/tests/test_views.py
```diff
@@ -215,18 +219,5 @@ def test_display_label(self):
...
-        assert speaker in response.context['current']
-
-    def test_display_label(self):
-        '''test display label for speaker'''
-        speaker = Person.objects.create(first_name='Jill', institution='IAS')
-        assert SpeakerListView().display_label(speaker) == \
-            speaker.institution
-
+def test_speakers_list_gone(client):
+    assert client.get('/people/speakers/').status_code == 410
```
thatbudakguy, line 223: oh nice, good call on this

thatbudakguy (approved):
lgtm; good call on the 410 gone for the speakers page. practically speaking, would that display to the user as our 404 page?

rlskoeser:
> lgtm; good call on the 410 gone for the speakers page. practically speaking, would that display to the user as our 404 page?

Good point! It doesn't right now, but it's possible people might land there. Let me fix that.

PR #282, opened 2021-02-11
fix/model permissions

thatbudakguy:
In order to make non-`Page`-derived models editable in the Wagtail admin, we needed to add permissions for them to the two main Wagtail groups (`Moderators` and `Editors`). This PR adds two migrations (one for people and one for projects) that take care of that.

It also adds a new method to the people exodus that:
- removes all superuser rights
- removes any user accounts that can't log in (`is_staff=False` or `is_active=False`)
- moves all remaining users to the Editors group

the hope is that by scaling all the permissions way back and calling this method early in the exodus, we'll know for sure if removing the permissions and users would impact any content (like blog posts), because it'll error. if the exodus runs without issue, we can later move people to the moderators group as determined by a charter, and make a subset of those people superuser. in the meantime, Editors should have permissions to do most day-to-day tasks on the site.

(my initial thought was to just make everyone who had superuser be a `Moderator`, but it turns out that's >80% of all currently active users! so, not very useful..."if everyone's super, no one will be.")


cdhweb/people/exodus.py
```diff
@@ -43,3 +48,53 @@ def people_exodus():
...
+    Historically, many users were given the is_superuser flag to avoid a bug
+    with Mezzanine permissions. In Wagtail, this is no longer necessary. In
+    addition, there are two new default groups: Moderators and Editors.
+
+    Users with is_staff or is_active set to False (preventing login) will be 
+    deleted, since there's no need to keep them in the new system. User accounts
+    are no longer associated with content (instead, People are).
+
+    Note that you will still need to assign users to the Moderators group and
+    turn on is_superuser manually after this runs. This can be done in console.
```
rlskoeser, line 70: `createcasuser` manage command with the `--admin` flag would work too

cdhweb/people/migrations/0026_modeladmin_perms.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # give moderators all permissions on Title and Person
+    moderators.permissions.set([
+        add_title, change_title, delete_title, view_title,
+        add_person, change_person, delete_person, view_person
+    ])
+
+    # editors have all permissions on Person, but none for Title
+    editors.permissions.set([
+        add_person, change_person, delete_person, view_person
```
rlskoeser, line 46: is view title permission needed to add it to a person, or only for the list view?

cdhweb/people/exodus.py
```diff
@@ -43,3 +48,53 @@ def people_exodus():
...
+        user.delete()
+    if non_staff:
+        logging.info("removed %d non-staff users" % non_staff)
+
+    # assign all remaining users to editors group; remove superuser.
+    for user in User.objects.all():
+        logging.debug("adding user %s to wagtail editors" % user)
+        editors.user_set.add(user)
+        user.is_superuser = False
+        user.save()
```
rlskoeser, line 99: Not that it matters for efficiency with the number of records we have, but you could use a queryset `update()` to set `is_superuser=False` on all users in a single line. I forget the syntax, but imagine there's a way to add a list of users to a group all at once.

cdhweb/people/exodus.py
```diff
@@ -43,3 +48,53 @@ def people_exodus():
...
+
+    # remove any groups that aren't editors or moderators - in cdhweb 2.x we
+    # had an unused "Content Editor" group that should be deleted
+    Group.objects.exclude(pk__in=(moderators.pk, editors.pk)).delete()
+
+    # remove any users who had login disabled for some reason, except the
+    # script user
+    non_staff = 0
+    for user in User.objects.filter(Q(is_staff=False) | Q(is_active=False)) \
+                            .exclude(settings["SCRIPT_USERNAME"]):
```
rlskoeser, line 87: Should be `settings.SCRIPT_USERNAME`

rlskoeser (commented):
Changes look reasonable to me. Documentation is very thorough and helpful.

I tried to run it locally and hit some errors, maybe related to the problems causing the tests to fail.

cdhweb/people/exodus.py
thatbudakguy: gonna keep it this way so that we can still log each user added, if that's ok

cdhweb/people/migrations/0026_modeladmin_perms.py
thatbudakguy: it's not needed to add it to a person, and removing it means that they don't see the heading for the list view in admin, which i think is what we want

PR #283, opened 2021-02-22
More person list revisions

rlskoeser:
Adjustments based on person list testing feedback — ref #267 

- `profile_url` logic didn't handle the case where there was no profile; no unit test, either so I added one
- related link display was due to bad variable in the template; I didn't see an obvious place to add a unit test for this (no template / display tests included in person page tests currently)
- affiliates was sorting on user last name, which doesn't exist anymore! corrected to use person last name

### DIFF
--- a/cdhweb/people/templates/people/profile.html
+++ b/cdhweb/people/templates/people/profile.html
@@ -20,7 +20,7 @@
 <div class="links">
     {# NOTE equiv to contributors in project #}
     <ul>
-    {% for link in person.related_links.all %}
+    {% for link in page.person.related_links.all %}
         <li><a href="{{ link.url }}">{{ link.type.name }}</a></li>
     {% endfor %}
     </ul>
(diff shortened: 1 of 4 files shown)


cdhweb/people/templates/people/profile.html
```diff
@@ -20,7 +20,7 @@ <h1>{{ page.title }}</h1>
 <div class="links">
     {# NOTE equiv to contributors in project #}
     <ul>
-    {% for link in person.related_links.all %}
+    {% for link in page.person.related_links.all %}
```
thatbudakguy, line 23: i figured out you can actually set `context_object_name` on wagtail pages after all; we should probably do that so this is `profile.person.related_links`, etc. I have set it for `Event` and `Project` in my branch for events

thatbudakguy (approved):
changes look good!

cdhweb/people/templates/people/profile.html
rlskoeser: do you want me to make that change now? or would it make sense to fold it into your branch?

thatbudakguy: probably more sense in my branch tbh. i'll just rebase as soon as we merge this and then make the change

Issue #252, opened 2020-12-08
As a moderator, I want existing events migrated to wagtail so that I can manage them in the new system.


thatbudakguy:
# testing notes
companion to #182; you may want to do that first to get a sense of how events look in the new system.
## as a user
- visit some event pages on the new site. does everything look similar to the corresponding project pages on the old site? try different kinds of events, like lectures, workshops, etc.
- remember that the test site will often be out-of-date with production, so any changes made in the last few weeks may not show in test. that's OK; you're looking for things that seem _wrong_ not things that are missing.
- note anything that looks off and leave a comment below.
# dev notes
depends on #184

thatbudakguy:
reducing to a 1 since exodus could be adapted directly from projects, with some simplifications

manawinters:
- The dropdown option does not work in the test site
- Names contain middle initials in the production, but not test site (eg. [Grant R. Wythoff](https://cdh.princeton.edu/events/2021/01/humanities-data-teaching-fellows-information-session/) in the Humanities Data Teaching Fellows Info Session)
- Otherwise events look the same on production and test sites

thatbudakguy:
@manawinters i think the dropdown not working is actually because the "this is a test site" banner is in the way! can you try clicking OK on the banner and seeing if the dropdown works after that? it worked for me.

the name discrepancy is OK; I think we only want to use the profile title (the one with the middle initial) on the profile going forward. please close this if you are able to operate the dropdown after clicking OK on the test banner.

Issue #184, opened 2019-12-20
As a moderator, I want to create and edit events in wagtail so that I can manage them just like other pages.


thatbudakguy:
# testing notes
## as a moderator
to test this, log into the new admin site (`/cms`) and check to see if you have a menu on the left side called "Events".

1. Make a new event by going to the "Events" menu, choosing "Event", and clicking "new event" at the top right. Feel free to also give the event an image/thumbnail and add other details, including speakers if you want.
2. Visit the event's page by clicking "preview" at the bottom. Check that the event page is visually similar to current CDH event pages. The picture displayed should be the main image you uploaded to the project (not the thumbnail). The page should display the detail you added to the event, including speakers and other data.
3. Try clicking the "export to iCal" button on the left and see if your browser downloads a file (it may offer to open it with your calendar app). this should work the same way as in production. if you download the event, check that the event details on your calendar match the ones on the test site.

moderators can also manage the list of locations and available event types.

1. Create a new event type from the "Events" menu on the left.
2. Give the project that you created above this new event type.
3. Check the project page you created above and see if the event type has been updated.
4. Create a new location from the "Events" menu on the left.
5. Add the location to the event you created above like you did in step 2.
6. Check the project page again and see if your location is listed.
# dev notes
`EventIndexPage` is notable because it uses by-semester view filtering logic, which we haven't done on the wagtail side before. we can try using wagtail's `RoutablePageMixin` and if that fails fall back to the s&co implementation.
## models
- [x] `Event` (convert to Wagtail page model, change image fields, add tags).
- [x] `EventTypeManager` (minimal changes)
- [x] `EventType` (minimal changes)
- [x] `EventQuerySet` (extends `PageQuerySet`)
## admin
- [x] `EventType`, uses modeladmin, editable by Moderator only
- [x] `Event`, uses modeladmin, remove from page tree, editable by all users
- [x] `Location`, uses modeladmin, editable by all users?

thatbudakguy:
bumping to 3 to account for new tests and possibly views

thatbudakguy:
bumping to 5 to account for testing (by proxy) `LastModifiedMixin`

thatbudakguy:
moved back to a 3; split off views to #286

manawinters:
@thatbudakguy , I tried to create a new event (step 1), and got the 500/chicken error message.

thatbudakguy:
[...]

PR #284, opened 2021-02-23
Events exodus

thatbudakguy:
- Create new Event and associated models (#184)
- Fallback strategy for importing large fixtures
- Fix homepage tests for project and event cards
- Better context object naming for some Page models
- Events exodus (#252)
- Set up permissions for EventType and Location
- Fix project timestamp field in export
- Admin config for Events models (#184)
- Use DetailView to route Event pages


cdhweb/events/models.py
```diff
@@ -215,3 +231,177 @@ def ical_event(self):
...
+class Event(Page, ClusterableModel):
+    """Page type for an event, such as a workshop, lecture, or conference."""
+
+    content = StreamField(BodyContentBlock, blank=True)
+    sponsor = models.CharField(max_length=80, null=True, blank=True)
+    start_time = models.DateTimeField()
+    end_time = models.DateTimeField()
+    location = models.ForeignKey(Location, null=True, blank=True,
+                                 on_delete=models.SET_NULL)
+    type = models.ForeignKey(EventType, null=True, on_delete=models.SET_NULL)
```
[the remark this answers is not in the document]
rlskoeser: oh, nm, I see the wagtail validation commments

cdhweb/events/templates/events/snippets/event_card.html
```diff
@@ -1,9 +1,10 @@
+{% load wagtailcore_tags wagtailimages_tags %}
 {# display a single event in card format #}
 {# FIXME: significant overlap with event_detail, esp. for schema properties #}
 <div class="card event" typeof="schema:Event">
-    <a property="schema:url" href="{{ event.get_absolute_url }}">
-        <div class="event-type">{{ event.event_type }}</div>
-        <div class="image" {% if event.thumb %}style="background-image:url('{{ MEDIA_URL }}{{ event.thumb }}')"{% endif %}></div>
+    <a property="schema:url" href="{{ event.get_url }}">
+        <div class="event-type">{{ event.type }}</div>
+        <div class="image" {% if event.thumbnail %}style="background-image:url('{{ MEDIA_URL }}{{ event.thumb }}')"{% endif %}></div>
```
rlskoeser, line 7: You're using `thumbnail` in the test but outputting `thumb`; maybe this is why images aren't showing up?

cdhweb/events/models.py
```diff
@@ -215,3 +231,177 @@ def ical_event(self):
...
+        FieldPanel("tags")
+    ]
+
+    # custom manager/queryset logic
+    objects = EventManager()
+
+    context_object_name = "event"
+
+    class Meta:
+        ordering = ("-start_time",)
```
rlskoeser, line 304: would it make sense to add a unique constraint here with a condition so it only applies to events with same month/year? Maybe it should be out of scope, but I think I remember you mentioning this was a bug and I think it might not be that hard see https://docs.djangoproject.com/en/3.1/ref/models/constraints/#uniqueconstraint

cdhweb/projects/models.py
```diff
@@ -235,6 +236,8 @@ class Project(Page, ClusterableModel):
     # custom manager/queryset logic
     objects = ProjectManager()
 
+    context_object_name = "project"
```
rlskoeser, line 239: 👍

cdhweb/events/models.py
rlskoeser: found a better option, what about using `unique_for_month` and `unique_for_year` on the slug field? 
https://docs.djangoproject.com/en/2.2/ref/models/fields/#unique-for-month

(do we inherit slug from `Page`? wonder if we can re-declare & override it.)

cdhweb/conftest.py
```diff
@@ -1,3 +1,8 @@
 """Fixtures/utilities that should be globally available for testing."""
 from cdhweb.pages.tests.conftest import site, homepage
-from cdhweb.projects.tests.conftest import projects_link_page
+# FIXME not sure how else to share fixtures that depend on other fixtures
+# between modules - if you import just the top-level fixture (e.g. "events"),
+# it fails to find the fixture dependencies, and so on all the way down. For
+# now this does what we want, although it pollutes the namespace somewhat
```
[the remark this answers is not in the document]
thatbudakguy: I couldn't find anything in the docs describing a situation where a fixture is shared across modules, rather than just across multiple test files in a module. The closest thing seems to be changing fixture scope, but I still want these to be created on every function call rather than shared :/

cdhweb/events/templates/events/snippets/event_card.html
thatbudakguy: 🤦🏻 good catch, thanks!

cdhweb/events/models.py
thatbudakguy: oooh, that's really nice. do we need both? i can try redeclaring slug and see what happens; we do inherit it from `Page`

thatbudakguy: nvm I guess we [can't do that](https://docs.djangoproject.com/en/3.1/topics/db/models/#field-name-hiding-is-not-permitted). we could try modifying _attributes_ of the field, although idk what would happen.

thatbudakguy: actually, we might not need to do anything at all - I think wagtail already validates uniquness for slugs at a particular location, and we're forcing all events to exist at the same point in the page tree, so creating a duplicate slug should be impossible

thatbudakguy: confirmed this is the case; wagtail won't let you save it with a duplicate slug even if the dates are different.

rlskoeser: ok — that seems like it's actually more uniqueness than we actually need (and I wonder if it might lead to weird slugs eventually), but better than allowing url collisions

thatbudakguy:
ok, here's the status of the tests.
- `TestUpcomingEventsView.test_last_modified`: usually passes, but occasionally fails because the timestamp is off somehow.
- `TestUpcomingEventsView.test_not_modified`: usually passes, but occasionally fails because it does not return a 304 Not Modified (perhaps same problem as `test_last_modified`).
- `TestUpcomingEventsView.test_head`: identical to `test_not_modified` but it's a HEAD request; presumably fails for the same reason (but at different times!)
- `TestEventRedirectView.test_event_redirect_by_id`: fails for unknown reasons when looking up the PK of the event; can't `reverse()` its URL.

I can't get any of these tests to reliably fail when run individually. Some (but not all) of them will fail each time _all_ tests are run, but I can't find a pattern of which will fail.

Issue #250, opened 2020-12-08
As an editor, I want existing blog posts migrated to wagtail so that I can manage them in the new system.


thatbudakguy:
# testing notes
this can be done in tandem with or as part of #287 - we just want to check that all the old blogposts were correctly moved to the new system (just like projects, events, etc.). remember that newer content/images will be missing because the data in the test site is old, and also the note on #287 about posts with no author.

# dev notes
depends on #185

Issue #185, opened 2019-12-20
As an editor, I want to create and edit blog posts in wagtail so that I can take advantage of its editing capabilities.


thatbudakguy:
# testing notes
there should now be a "blog posts" menu in the admin, and you should be able to create new blog posts. try adding images and other content to a blogpost. check that:

- if you publish a post and set it as featured, it shows up on the homepage
- draft posts don't appear on the homepage or anywhere else
- you can add multiple authors to a blog post, and the order you add them in is also the order they're listed in

after you create a post, visit its page on the site and compare it visually to production. in particular, look for:

- the next/previous links at the bottom of the page should point to the correct posts in order by date
- the authors should be listed correctly, in order
- any special content in the post, like rich text or a video, should look the same way as it does in production

# dev notes
very similar to Events app, but the `BlogPostIndexPage` uses the same year/month routing that we have on s&co. `BlogPost` needs to update its fields to use wagtail versions and remove some unused stuff like "related posts". we can also turn on tags for it.
## models
- [x] `BlogPost` (becomes wagtail page, remove related posts, revisit truncation logic)
- [x] `BlogPostQuerySet` (update `recent()` method)
- [x] `BlogPostIndexPage` (new page model that uses current view logic)
## admin
- [x] `BlogPost` modeladmin

PR #289, opened 2021-03-02
blog exodus

thatbudakguy:
No tests yet; opening as draft to confirm changes look good.

- Add/update Blog models (#185)
- Set up admin for Blog posts
- Remove mezzanine blog sitemap & admin
- Create blogpost exodus (#250)


cdhweb/blog/models.py
```diff
@@ -99,4 +107,96 @@ def get_absolute_url(self):
...
+
+    context_object_name = "post"
+
+    class Meta:
+        ordering = ("-first_published_at",)
+
+    @property
+    def short_title(self):
+        """Shorter title with ellipsis."""
+        return Truncator(self.title).chars(65)
```
rlskoeser, line 165: `Truncator` seems like a good thing to know about!

cdhweb/blog/models.py
```diff
@@ -99,4 +107,96 @@ def get_absolute_url(self):
...
+class BlogPostTag(TaggedItemBase):
+    """Tags for Blog posts."""
+    content_object = ParentalKey(
+        "blog.BlogPost", on_delete=models.CASCADE, related_name="tagged_items")
+
+
+class BlogPost(Page, ClusterableModel):
+    """A Blog post, implemented as a Wagtail page."""
+
+    authors = models.ManyToManyField("people.Person", related_name="posts")
```
rlskoeser, line 127: Can we easily make author an ordered many to many with wagtail? Probably should be out of scope if it's not easy, but thought I would ask. (It's something we've wanted before)

rlskoeser (commented):
Everything here looks reasonable in terms of approach. Looks like you've been able to adapt the approach you figured out for other apps, which is great to see.

cdhweb/blog/models.py
thatbudakguy: Great q! I know wagtail has an `Ordered` mixin and this seems like a core use-case; maybe they have an example already. I'll look

thatbudakguy:
changes since last review:
- tests for more things (views, templates)
- updates to blog post detail template, homepage carousel template, others
- some convenience methods like `str()` for `Author` and `BlogPost`
- check for bad HTML during exodus and fix it using `BeautifulSoup`
- don't return person profile URLs if the profile is a draft
- add `PagePreviewDescriptionMixin` to `BlogPost`
- make sure `BlogPost` querysets can sort by date published via `recent()` since default ordering for pages isn't allowed

cdhweb/pages/exodus.py
```diff
@@ -26,7 +27,18 @@ def convert_slug(slug):
 
 def to_streamfield(content):
     """Output a json object for given richtext as a "migrated" streamfield."""
-    return json.dumps([{"type": "migrated", "value": content}])
+    # Wagtail chokes on unclosed <img> tags, throwing: "Unmatched tags:
+    # expected img, got figure". There may be other issues lurking in the html
+    # from mezzanine. Instead of trying to parse it with a regex, we let a real
+    # parser (BeautifulSoup) do the work.
+    pretty_content = BeautifulSoup(content, "lxml").prettify()
```
rlskoeser, line 34: nice solution

cdhweb/people/models.py
```diff
@@ -286,6 +287,12 @@ def current_title(self):
         if current_positions.exists():
             return current_positions.first().title
 
+    @property
+    def most_recent_title(self):
+        """Return the most recent of any titles held by this Person."""
+        if self.positions.exists():
+            return self.positions.first().title
```
rlskoeser, line 294: IDK how much it matters, but I think this will run two queries when we only need one. Slightly better to use `first()` instead of `exists()` and then if the first result is not none return the title attribute.

cdhweb/people/models.py
```diff
@@ -429,21 +436,12 @@ class Profile(Page):
...
-            context.update({
-                "opengraph_type": "profile",
-                "recent_posts": recent_posts
-            })
-
+        # get 3 most recent published posts with this person as author;
+        # add to context and set open graph metadata
+        context.update({
+            "opengraph_type": "profile",
+            "recent_posts": self.person.posts.live().recent()[:3]
```
rlskoeser, line 443: Nice! So much more readable

rlskoeser (changes requested):
Code changes since my last review look good. One small suggestion, but IDK how much it matters.

Maybe something that got missed from a previous round: I tried to use the `createcasuser` command and got an error, it looks like the extra init method was renamed (profile -> person) but not updated in settings.

Could we make the verbose label for blog `is_featured` to just `featured`? It looks/reads a little odd to me in the wagtail admin.

I don't love the `CDH Staff` default author, it doesn't seem meaningful to me. Is author required? Would it make more sense to not put one? (We may want to check schema.org to see if they think author is required; although IDK how useful it is to put a non-author author.). 

I also think it may be introducing some errors in the exodus — I was surprised to see how many recent posts were authored by "CDH Staff", and in particular noticed that the one on "New Frontiers in Slavic DH" was staff, which seemed wrong. In production, it's shown as being authored by Alexander Jacobson, so something is getting lost right now.  Is using CDH staff is a default author obscuring actual errors? What are the implications for not using it as a default?

When I went to the update list (because I wanted to compare authors with production), I got a template error: `'wagtailpages_tags' is not a registered tag library`; seeing the same error on profile pages that should be listing blog posts.

I also noticed that featured items on the homepage are only showing title and no blurb for the beginning of the text. I was surprised that my local exodus only resulted in one featured blog post, but I don't know if that's accurate or not (possibly I changed it when I was testing something the last time I did anything).

rlskoeser:
I also just got a field error when I tried to search the wagtail blog list

thatbudakguy:
- updated `PUCAS_LDAP.EXTRA_USER_INIT`
- renamed field to `BlogPost.featured`
- removed default authors; as far as I could tell schema.org doesn't require one and i couldn't think of any issues with not having one
- fixed wrong name for wagtail template lib (it should've been `wagtailcore_tags`)
- missing descriptions was due to the mixin that generates them not recognizing streamfield blocks of type `migrated` as valid body content to use; I elected to just copy over generated mezzanine descriptions as we have for other content as part of the exodus
- fixed wagtail admin config that was causing the field error

rlskoeser (commented):
The updates you've made look good, but I'm not able to run things locally due to some data/migration errors. I undid and redid blog migrations, but then trying to run the exodus generated an error — maybe because I haven't refreshed the database since the last time I ran the exodus script? Would it be helpful for me to refresh my data and run everything?

The thing I wanted to check, which I'm not able to right now, and which I'm still concerned about is the missing blog authors. Were you able to confirm if that's been resolved by the most recent changes?   If it isn't, I wondered if maybe your user cleanup is too aggressive and is removing non-active blog post authors (but I see that doesn't run until after the blog exodus, so maybe it can't be that).

thatbudakguy:
I confirmed that the missing author you found (alex on the slavic DH post) was resolved, and I'm not getting any exodus errors with a recent dbdump. I couldn't find any instances where authors were different between my local dev and production.

rlskoeser (approved):

PR #290, opened 2021-03-11
person email

thatbudakguy:
small PR based off of #289; will rebase against develop when that is merged.

fixes an issue identified by @manawinters on #181 – some people no longer have emails displayed on their `Profile` in v3 because that information was stored on `User` and would otherwise be lost (most extraneous `User`s will become `Person`s in v3).

this PR resolves it by:
- adding an `Email` field to `Person`
- populating the field if it's empty when `init_person_from_ldap` is run
- populating the field based on `User.email` in an existing migration so that information from `User` isn't lost

it also uses a better context object name in the `Profile` template and improves template tests.


rlskoeser:
@thatbudakguy it's hard to see the specific changes you want me to review. Can we either configure the PR to merge into the branch for #289 or wait until that's merged?

thatbudakguy:
yes, plan is to wait until #289 is merged

rlskoeser (approved):
All the changes here look reasonable to me.

PR #295, opened 2021-03-22
feature/sitemaps

thatbudakguy:
- Add wagtail default sitemap
- Add sitemap priority function for blog posts
- Add sitemap priority function for projects
- Set wagtail Site's hostname to cdh.princeton.edu in exodus

I implemented the sitemap priority calculation for projects and blog posts to match how they're calculated in current production; they became methods on wagtail `Page`. As far as I can tell, this is all we need to do! It looks like everything's present in the sitemap and automatically uses the right modification dates, etc.

### DIFF
--- a/cdhweb/blog/models.py
+++ b/cdhweb/blog/models.py
@@ -204,6 +204,15 @@ class BlogPost(BasePage, ClusterableModel, PagePreviewDescriptionMixin):
             })
             return site_id, root_url, page_path
 
+    def get_sitemap_urls(self, request):
+        """Override sitemap listings to add priority for featured posts."""
+        # output is a list of dict; there should only ever be one element. see:
+        # https://docs.wagtail.io/en/stable/reference/contrib/sitemaps.html#urls
+        urls = super().get_sitemap_urls(request=request)
+        if self.featured:
+            urls[0]["priority"] = 0.6   # default is 0.5; slight increase
+        return urls
+
 
 class BlogLinkPage(LinkPage):
     """Container page that defines where blog posts can be created."""
--- a/cdhweb/pages/management/commands/exodus.py
+++ b/cdhweb/pages/management/commands/exodus.py
@@ -105,6 +105,10 @@ class Command(BaseCommand):
         site.save()
         Page.objects.filter(depth=2).exclude(pk=homepage.pk).delete()
 
+        # set the site's hostname so that sitemaps will be valid
+        site.hostname = "cdh.princeton.edu"
+        site.save()
+
         # create a LinkPage to serve as the projects/ root (project list page)
         try:
             old_projects = MezzaninePage.objects.get(slug="projects")
--- a/cdhweb/projects/models.py
+++ b/cdhweb/projects/models.py
@@ -290,6 +290,17 @@ class Project(BasePage, ClusterableModel):
             .exclude(membership__in=self.current_memberships()) \
             .order_by("last_name")
 
+    def get_sitemap_urls(self, request):
+        """Override sitemap to prioritize projects built by CDH with a website."""
+        # output is a list of dict; there should only ever be one element. see:
+        # https://docs.wagtail.io/en/stable/reference/contrib/sitemaps.html#urls
+        urls = super().get_sitemap_urls(request=request)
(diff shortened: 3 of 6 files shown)


cdhweb/projects/models.py
```diff
@@ -290,6 +290,17 @@ def alums(self):
...
 
+    def get_sitemap_urls(self, request):
+        """Override sitemap to prioritize projects built by CDH with a website."""
+        # output is a list of dict; there should only ever be one element. see:
+        # https://docs.wagtail.io/en/stable/reference/contrib/sitemaps.html#urls
+        urls = super().get_sitemap_urls(request=request)
+        if self.website_url and self.cdh_built:
+            urls[0]["priority"] = 0.7
+        elif self.website_url or self.cdh_built:
+            urls[0]["priority"] = 0.6
```
rlskoeser, line 301: should we also add priority for things that are highlighted?

cdhweb/pages/management/commands/exodus.py
```diff
@@ -105,6 +105,10 @@ def handle(self, *args, **options):
         site.save()
         Page.objects.filter(depth=2).exclude(pk=homepage.pk).delete()
 
+        # set the site's hostname so that sitemaps will be valid
+        site.hostname = "cdh.princeton.edu"
```
rlskoeser, line 109: do we need to do anything for qa, or does it not matter?

cdhweb/blog/models.py
```diff
@@ -204,6 +204,15 @@ def get_url_parts(self, *args, **kwargs):
             })
             return site_id, root_url, page_path
 
+    def get_sitemap_urls(self, request):
+        """Override sitemap listings to add priority for featured posts."""
+        # output is a list of dict; there should only ever be one element. see:
+        # https://docs.wagtail.io/en/stable/reference/contrib/sitemaps.html#urls
+        urls = super().get_sitemap_urls(request=request)
```
rlskoeser, line 211: seems confusing that this returns a list but there's only ever 1 element, but I see that this is working

rlskoeser (approved):
Looks good!

cdhweb/blog/models.py
thatbudakguy: agree...I guess the use case is a single page that needs multiple sitemap entries, but having trouble imagining why you'd want that.

cdhweb/pages/management/commands/exodus.py
thatbudakguy: shouldn't matter since QA can't be crawled without VPN, i think

cdhweb/projects/models.py
thatbudakguy: that makes sense to me. should we also maybe prioritize staff profiles (vs e.g. grad fellows)? trying to think of other uses for it. if we do add more priority to projects so they can have up to 0.8, I wonder if we should give more to featured blogposts too so that they're not still at 0.6 relative to projects.

thatbudakguy: it might even just be simpler to only bump priority slightly for highlighted projects and remove the current logic...i think highlighted and cdh-built are pretty much 1:1 and i'm not sure how important website URL is

cdhweb/blog/models.py
rlskoeser: I can — if you generate different formats of the same page/view (similar to text/html/pdf for startwords)

cdhweb/projects/models.py
rlskoeser: What if we de-prioritize projects without a url? A lot of the older projects that never got far enough along to get a url are kind of vapour ware in a way. (But might not apply to working groups...)  

Not all cdh-built projects are highlighted (although agree there is substantial overlap; but see e.g. the New NLP project); it's more current highlights / current work, I think. 

I agree that setting them to 0.8 seems too high. Maybe we're overcomplicating it — we can set to current logic and make a chore to investigate revising, if you think it's worth a little more thought.

thatbudakguy: that sounds good to me – I'll make a chore and we can leave it this way for now. definitely seems like there are a lot of things to consider.

Issue #247, opened 2020-12-08
As an admin, I want to see the edit history for a page, including edits before the wagtail migration, so that I can see the history of edits to a page.


thatbudakguy:
# testing notes
this one is pretty simple – you should now be able to view the edit history for each page! it's located at the top under the title; you click the "History" text with the clock. try this for a bunch of different things - projects, profiles, regular pages, etc, etc. it won't work for things that aren't pages (e.g. membership) but it should work for everything else.

also check that you see an entry in the history marking the page's migration from cdhweb v2. we want to mark this so that we know when pages were moved, or if they are new for v3 and didn't get moved.

lastly, you should see when you log in to wagtail a list of the most recent changes you made (using the outdated test data, of course).

# dev notes
we want to:
- [x] transfer django admin log entries from old page object to new (update both contenttype and foreign key)
- [x] see if django admin "view history" still works after this
- [x] make sure the exodus to wagtail event itself is recorded, by a script user

thatbudakguy:
bumping to 3 since this required learning how to use `PageLogEntry` and wagtail's audit log features.

PR #298, opened 2021-03-23
feature/exodus history (#247)

thatbudakguy:
- Implement django log entry exodizing
- Add a custom log action type for exodus
- Exodize log entries in main exodus script
- Exodize log entries for blog, event, people, project pages
- Update user/group exodus

this uses the new strategy to migrate log entries to `PageLogEntry`. It also makes sure an initial page revision exists in wagtail, otherwise the "history" button doesn't show up until you make some kind of change. you can access the history for each page at the top next to the "last revision published at" display, or via the global log (that one is in "reports" on the left menu and is only visible to superusers). all of the `log_action=False` changes are to prevent populating the logs with changes that are made via the exodus; instead we add a single custom log entry per model to mark the exodus event.

for the user/group exodus, we now delete accounts with no login/edit history (the vast majority) and mark any remaining accounts with no current cdh position as inactive (this seemed to do the right thing when testing). also adds rsk and nb as superusers.

### DIFF
--- a/cdhweb/pages/exodus.py
+++ b/cdhweb/pages/exodus.py
@@ -114,19 +114,23 @@ def exodize_history(mezzanine_page, wagtail_page):
                                         action="wagtail.create")
     creation.delete()
 
-    # get all the old admin log entries for this page. in some cases (seems to
-    # happen for regular content pages) there will be none; in this case we
-    # will end up creating a new revision with a timestamp of right now, and
-    # the only log entry will be marking the exodus
-    ctype = ContentType.objects.get_for_model(mezzanine_page)
+    # get all the old admin log entries for this page. content pages will
+    # have entries associated with their most specific subclass, but other
+    # page models won't, so we just use get_for_model directly. save the 
+    # user and timestamp from the most recent edit for later.
+    try:
+        specific_model = getattr(mezzanine_page, mezzanine_page.content_model)
+    except AttributeError:
+        specific_model = mezzanine_page
+    ctype = ContentType.objects.get_for_model(specific_model)
     entries = LogEntry.objects.filter(object_id=mezzanine_page.pk,
                                       content_type_id=ctype.pk)
     User = get_user_model()
     if entries.exists():
         logging.debug("exodizing %d log entries for %s" %
                       (entries.count(), mezzanine_page))
-        last_user = User.objects.get(pk=entries.last().user_id)
-        last_edit = entries.last().action_time
+        last_user = User.objects.get(pk=entries.first().user_id)
+        last_edit = entries.first().action_time
     else:
         last_user = None
         last_edit = timezone.now()
--- a/cdhweb/pages/exodus.py
+++ b/cdhweb/pages/exodus.py
@@ -3,16 +3,27 @@
 import json
 import logging
 import os
+from json.decoder import JSONDecodeError
 
 from bs4 import BeautifulSoup
+from django.conf import settings
+from django.contrib.admin.models import ADDITION, CHANGE, DELETION, LogEntry
+from django.contrib.auth import get_user_model
+from django.contrib.contenttypes.models import ContentType
+from django.utils import timezone
 from django.utils.html import strip_tags
-from wagtail.documents.blocks import DocumentChooserBlock
+from wagtail.core.models import PageLogEntry
 from wagtail.images.models import Image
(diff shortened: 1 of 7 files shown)


cdhweb/pages/exodus.py
```diff
@@ -88,7 +99,83 @@ def exodize_attachments(mezzanine_page, wagtail_page):
...
+    # make sure the page has only one real creation event; we want to use
+    # the date it was created in mezzanine
+    creation = PageLogEntry.objects.get(page=wagtail_page,
+                                        action="wagtail.create")
+    creation.delete()
+
+    # get all the old admin log entries for this page. in some cases (seems to
+    # happen for regular content pages) there will be none; in this case we
+    # will end up creating a new revision with a timestamp of right now, and
+    # the only log entry will be marking the exodus
```
rlskoeser, line 120: Is it possible that you're looking at the wrong object? i.e. the log entries are associated with the base page object and not the content page?
(Could also be possible they don't exist!)

cdhweb/pages/exodus.py
```diff
@@ -88,7 +99,83 @@ def exodize_attachments(mezzanine_page, wagtail_page):
...
+    # for the definition of the custom "exodus" action.
+    PageLogEntry.objects.log_action(
+        instance=wagtail_page,
+        data={
+            "old_pk": mezzanine_page.pk,
+            "old_slug": mezzanine_page.slug,
+            "old_title": mezzanine_page.title
+        },
+        action="cdhweb.exodus",
+        user=script_user
```
rlskoeser, line 177: cool!

cdhweb/pages/management/commands/exodus.py
```diff
@@ -95,15 +97,24 @@ def handle(self, *args, **options):
...
         site.save()
-        Page.objects.filter(depth=2).exclude(pk=homepage.pk).delete()
+        Page.objects.filter(depth=2) \
+                    .exclude(pk=homepage.pk) \
+                    .delete()
+
+        # also delete any log entries referencing pages from previous runs, 
+        # otherwise they'll stick around and point to nothing since PKs change.
+        # objects.delete() isn't implemented so we do it via iterating
+        for entry in PageLogEntry.objects.all():
```
rlskoeser, line 116: You should be able to just do `.all().delete()` (delete isn't implemented on the manager, but it is available on the queryset)

cdhweb/people/exodus.py
```diff
@@ -66,24 +67,27 @@ def user_group_exodus():
     """Convert groups from Mezzanine configuration to Wagtail configuration.
 
     - Removes any groups outside Wagtail default Moderators and Editors
-    - Removes all users with the is_staff flag set to False
-    - Removes all users with the is_active flag set to False
+    - Removes all users who have never logged in/made a change to the site
     - Moves all remaining users into the Moderators group
+    - Marks any users without a current CDH position as inactive (no login)
+    - Whitelists RSK and NB as superusers
```
rlskoeser, line 73: thank you 🙏

cdhweb/people/exodus.py
```diff
@@ -98,21 +102,38 @@ def user_group_exodus():
...
         logging.debug("adding user %s to wagtail moderators" % user)
         moderators.user_set.add(user)
         user.is_superuser = False
         user.save()
+        try:
+            person = Person.objects.get(user=user)
+            if not person.current_title:
+                logging.debug("marking user %s non-active" % user)
+                user.is_active = False
+                user.save()
```
rlskoeser, line 125: Seems like this might be possible to do in a queryset filter and update, but I guess it's valuable to see explicitly what's being done per user

rlskoeser (approved):
This is so cool! Working really well, I love seeing the exodus migration log in the history.

I'm now seeing a page with my recent changes, which I don't think was showing up before — it must be based on these logs. (Am also shocked at how many times I apparently save my blog post drafts when I'm working on them...)

It looks like this is only working for pages, can the same logic be applied to other models that are managed in wagtail now? (people, memberships, roles, grant types, etc, etc). What about images and documents, does wagtail support displaying history for them also?

cdhweb/pages/management/commands/exodus.py
thatbudakguy: 💡 thank you!!

PR #300, opened 2021-03-24
Wagtail implementation for site navigation

rlskoeser:
- update exodus code to set wagtail `show_in_menus` based on mezzanine main nav/ footer nav menu status (or hard-coded for cases known to be needed)
- revise exodus & page models for projects to rearrange hierarchy so landing page is the parent of the section
- add logic to adjust top-level page order after exodus completes so that wagtail page order matches current nav order
- add a context processor to put the wagtail site into context for generating menus (couldn't figure out if the wagtail settings was supposed to provide this for us; I couldn't get it to give me the  site object)
- revise main navigation and footer menu templates to use wagtail pages queryset and `show_in_menus` configuration
- revise child page navigation to show all child pages except for profile and project pages (no menu configuration)

## notes
- menus don't currently filter on public/private; should we? (I don't think we're using private pages anywhere now; do we think we might start?)
- have updated existing tests but not added any new ones; advice appreciated on what you think is important to test (I'll check the coverage report also)


cdhweb/pages/context_processors.py
```diff
@@ -14,3 +16,7 @@ def page_intro(request):
     if page_intro:
         return {'page_intro': page_intro}
     return {}
+
+
+def wagtail_site(request):
+    return {'site': Site.find_for_request(request)}
```
thatbudakguy, line 22: fyi there's also a top-level context processor in `cdhweb/context_processors.py` that sets `site` to the _django_ site object. not sure how or if that interacts with this

cdhweb/pages/management/commands/exodus.py
```diff
@@ -429,3 +449,19 @@ def image_exodus(self):
...
+    ]
+
+    def adjust_toplevel_order(self, homepage):
+        # find and move each child page to first child
+        # in reverse of desired order; any pages not in our
+        # list should be ignored and end up at the end
+        for slug in reversed(self.nav_order):
+            page_to_move = homepage.get_children().get(slug=slug)
+            # make this page the first child
+            page_to_move.move(homepage, 'first-child')
```
thatbudakguy, line 467: i think `.move()` is an event that is stored in the wagtail audit log, so for consistency with stuff i did in the history PR/story you might want `.move(log_action=False)` to prevent this showing up in the history

cdhweb/projects/exodus.py
```diff
@@ -4,15 +4,15 @@
...
+from cdhweb.projects.models import OldProject, Project, ProjectsLandingPage
 
 
 def project_exodus():
     """exodize all project models"""
     # get the top-level projects link page
     try:
-        project_link = ProjectsLinkPage.objects.get()
-    except ProjectsLinkPage.DoesNotExist:
+        project_link = ProjectsLandingPage.objects.get()
```
thatbudakguy, line 14: could update comment and variable name for consistency, but not that important

thatbudakguy (approved):
everything looks good! re:testing, the only thing that seems easy enough and worth the time is the `child_page.html` template – i think writing a test for that wouldn't be too tough, but the footer and main nav are probably complex enough to require fixtures to really test, and I don't think we have time/necessarily need that right now.

cdhweb/pages/context_processors.py
rlskoeser: Whoa, good catch. I can't remember what we use that for.. I wonder if we still need both, or if the wagtail site is compatible enough with the base django site to use it instead... Should I rename the new one to wagtail site? Also, what do you think about including it in the top level context processors instead of adding a new one?

cdhweb/pages/management/commands/exodus.py
rlskoeser: ok, thanks for flagging

cdhweb/projects/exodus.py
rlskoeser: oh, thanks — I did try to update the names, I missed this one

cdhweb/pages/context_processors.py
thatbudakguy: One thing I did see it used for was in the `robots.txt` like we were talking about this morning...presumably the wagtail site should work the same? I'm fine with just replacing the method in the top-level context processor to return the wagtail site. It seems like the only thing we'd need the django one for is its hostname, which the wagtail one has.

thatbudakguy (approved):
changes look good!

PR #301, opened 2021-03-24
feature/head meta

thatbudakguy:
- Use "page" context object naming everywhere for template compatibility
- Update <title> elements site-wide for brevity
- Update all <head> metadata, including keywords
- Exodize tags on projects, events, and blog posts so they can be used as keywords

### DIFF
--- a/cdhweb/blog/exodus.py
+++ b/cdhweb/blog/exodus.py
@@ -2,10 +2,13 @@
 import logging
 
 from mezzanine.core.models import CONTENT_STATUS_PUBLISHED
+from taggit.models import Tag
 
-from cdhweb.blog.models import Author, BlogLinkPage, BlogPost, OldBlogPost
-from cdhweb.pages.exodus import (convert_slug, exodize_attachments, exodize_history,
-                                 get_wagtail_image, to_streamfield)
+from cdhweb.blog.models import (Author, BlogLinkPage, BlogPost, BlogPostTag,
+                                OldBlogPost)
+from cdhweb.pages.exodus import (convert_slug, exodize_attachments,
+                                 exodize_history, get_wagtail_image,
+                                 to_streamfield)
 from cdhweb.people.models import Person
 
 
@@ -58,4 +61,8 @@ def blog_exodus():
         exodize_attachments(post, post_page)
         exodize_history(post, post_page)
 
-        # NOTE no tags to migrate
+        # exodize keywords as tags
+        for kw in post.keywords.all():
+            logging.debug("exodizing blog tag %s" % kw.keyword.title)
+            tag = Tag.objects.get_or_create(name=kw.keyword.title)[0]
+            BlogPostTag.objects.create(content_object=post_page, tag=tag)
--- a/cdhweb/blog/templates/blog/blogpost_detail.html
+++ b/cdhweb/blog/templates/blog/blogpost_detail.html
@@ -2,14 +2,14 @@
 {% load wagtailcore_tags humanize %}
 {# simple blog post page; adapted in part from mezzanine blog post template #}
 
-{% block page-subtitle %}{{ post.title }} | {{ block.super }}{% endblock %}
+{% block page-subtitle %}{{ page.title }} | {{ block.super }}{% endblock %}
 
 {% block headattrs %} xmlns:article="http://ogp.me/ns/article#"{% endblock %}
 {% block extra-head %}
-    <meta name="article.published_time" content="{{ post.first_published_at.isoformat }}"/>
-    <meta name="article.modified_time" content="{{ post.last_published_at.isoformat }}"/>
-    {% if post.authors.exists %}
-      {% for author in post.authors.all %}
+    <meta name="article.published_time" content="{{ page.first_published_at.isoformat }}"/>
+    <meta name="article.modified_time" content="{{ page.last_published_at.isoformat }}"/>
+    {% if page.authors.exists %}
+      {% for author in page.authors.all %}
         {% if author.person.profile_url %}
         {# facebook requires url; provide local or external if we have it #}
     <meta name="article.author" content="{{ author.person.profile_url }}"/>
@@ -25,16 +25,16 @@
 {% block main %}
 <article typeof="schema:BlogPosting" class="blogpost">
     <header>
-        <h1 property="schema:headline">{{ post.title }}</h1>
-        <meta property="schema:url schema:mainEntityOfPage" content="{% pageurl post %}"/>
+        <h1 property="schema:headline">{{ page.title }}</h1>
+        <meta property="schema:url schema:mainEntityOfPage" content="{% pageurl page %}"/>
         {# FIXME: do we need full url here? #}
-        <meta property="schema:datePublished" content="{{ post.first_published_at|date:"c" }}"/>
-        <meta property="schema:dateModified" content="{{ post.last_published_at|date:"c" }}"/>
+        <meta property="schema:datePublished" content="{{ page.first_published_at|date:"c" }}"/>
+        <meta property="schema:dateModified" content="{{ page.last_published_at|date:"c" }}"/>
 
         <div class="byline">
-        {% if post.authors.exists %}
+        {% if page.authors.exists %}
         <div class="authors">
-        {% for author in post.authors.all %}
+        {% for author in page.authors.all %}
             <span property="schema:author" typeof="schema:Person">
                 <a {% if author.person.profile_url %}property="schema:url" href="{{ author.person.profile_url }}"{% endif %}>
                 <span property="schema:name">{{ author.person }}</span></a>{% if not forloop.last %}, {% endif %}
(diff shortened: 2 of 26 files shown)


cdhweb/pages/templates/cdhpages/home_page.html
```diff
@@ -1,7 +1,7 @@
 {% extends 'base.html' %}
 {% load wagtailcore_tags %}
 
-{% block page-title %}{{ page.seo_title }}{% endblock %}
+{% block page-title %}Home{% endblock %}
```
rlskoeser, line 4: I don't think "Home" is actually a meaningful title, esp. out of context. I also think it would be better if this is user-editable vs hardcoded in the template

cdhweb/projects/templates/projects/project_list.html
```diff
@@ -1,6 +1,6 @@
 {% extends "base.html" %}
 
-{% block page-subtitle %}{{ title }} | {% endblock %}
+{% block page-title %}{{ title|default:'People' }}{% endblock %}
```
rlskoeser, line 3: Do we need a default? And if so, this seems like the wrong default

templates/base.html
```diff
@@ -15,11 +15,11 @@
       gtag('config', 'UA-87887700-1', {'anonymize_ip': true});
     </script>
     {% endif %}
-    <title>{% block page-title %}{% block page-subtitle %}{% if page %}{{ page.meta_title|add:" | "}}{% endif %}{% endblock %}The Center for Digital Humanities at Princeton{% endblock %}</title>
+    <title>{% block page-title %}{% if page %}{{ page.seo_title }}{% endif %}{% endblock %} – CDH@Princeton</title>
```
rlskoeser, line 18: Ah, this is nice to handle in one place. Is that why all the others are being removed? If so, that's great.

Should this not be `page.seo_title` if set, otherwise `page.title` ? Suspect that most of our pages do _not_ have an seo specific title set.

Also, should the CDH@Princeton be overridable? seems like you wouldn't want it included on e.g. the home page. (Although that might be the _only_ place where it matters... 🤔 )

templates/snippets/head_meta.html
```diff
@@ -1,20 +1,20 @@
...
-    <meta property="og:description" content="{{ page.description }}"/>
-    <meta property="og:site_name" content="Center for Digital Humanities @ Princeton University"/>
-    {# twitter card #}
-    <meta name="twitter:card" content="{{ twitter_card_type|default:'summary' }}" />
-    <meta name="twitter:site" content="@PrincetonDH" />
-    <meta name="twitter:title" content="{{ page.meta_title }}"/>
-    <meta name="twitter:description" content="{{ page.description }}" />
-    <meta name="twitter:image" content="{% firstof twitter_image meta_preview_image %}" />
+<meta name="keywords" content="{{ page.tags.all|join:', ' }}"/>
+<meta name="description" content="{{ page.search_description }}"/>
```
rlskoeser, line 4: suspect this might need to be escaped

templates/site_index.html
```diff
@@ -1,7 +1,7 @@
 {% extends 'base.html' %}
 {% load mezzanine_tags wagtailcore_tags %}
 
-{% block page-title %}{% if page %}{{ page.meta_title }}{% else %}The Center for Digital Humanities at Princeton{% endif %}{% endblock %}
+{% block page-title %}Home{% endblock %}
```
rlskoeser, line 4: What's the difference between this template and the cdhpages home page template? Is this one no longer in use?

cdhweb/pages/templates/cdhpages/home_page.html
rlskoeser: Would like see the homepage keep the full name similar to what it has now (Center for Digital Humanities at Princeton)

templates/base.html
rlskoeser: And... I know you want to keep this short, but I'm not convinced CDH@Princeton is meaningful/recognizable enough. Is it really so terrible to keep the site name spelled out? Can't people with screen readers skip over it 
 once they hear it starts repeating?

cdhweb/blog/exodus.py
```diff
@@ -57,4 +59,8 @@ def blog_exodus():
         # transfer attachments
         exodize_attachments(post, post_page)
 
-        # NOTE no tags to migrate
+        # exodize keywords as tags
+        for kw in post.keywords.all():
+            logging.debug("exodizing blog tag %s" % kw.keyword.title)
+            tag = Tag.objects.get_or_create(name=kw.keyword.title)[0]
+            BlogPostTag.objects.create(content_object=post_page, tag=tag)
```
rlskoeser, line 68: I'm surprised you have to do it this way — if it works the way taggit normally does, you ought to be able to just use `post.tags.add` with the list. (Not that you need to change it for this, since it's a one-time thing and clearly it works; just sharing as an FYI)

templates/snippets/head_meta.html
rlskoeser, line 4: keywords and description are being included even if they're empty

templates/snippets/head_meta.html
```diff
@@ -1,20 +1,20 @@
...
+{# determine preview image: specified by page, cdh icon based on url, or default image #}
+{% firstof preview_image page.get_url|url_to_icon_path default_preview_image as meta_preview_image %}
+{# open graph metadata #}
+<meta property="og:title" content="{{ page.seo_title }}" />
```
rlskoeser, line 8: I checked a little bit, title is empty for all the pages I looked at. I think this needs to be something like `{% firstof page.seo_title page.title page_title %}` — I checked, that's the context variable set in non-page views like the project list pages. Maybe set this once similar to the preview image.

rlskoeser (changes requested):
Changes look good generally. It's too bad we have to lose the context-specific names for event, post, etc.; wish there was a way to alias them so we could have them as both event and page. But seems best for now to keep them as page for title /meta display.

Few small things that I think still need to be addressed.

cdhweb/pages/templates/cdhpages/home_page.html
thatbudakguy: i think that's the current `seo_title`, so I'll just revert this

cdhweb/projects/templates/projects/project_list.html
thatbudakguy: good catch!! the other list templates had defaults, so I put one here, but obvi didn't edit after pasting

templates/base.html
thatbudakguy: The spelled-out version is 47 characters, which is ~60% of our allowed title space. That seems like too much real estate to give up on every page, especially if Google may just add it to search results on its own anyway.

templates/site_index.html
thatbudakguy: yup – I edited both for consistency but this one won't be used anymore.

templates/base.html
rlskoeser: ok, let's try it. we can always tweak if we need to

cdhweb/blog/exodus.py
thatbudakguy: again, just doing it for logging purposes. probably more verbosity than we need, but speed isn't really a concern, and I like to read everything!

templates/snippets/head_meta.html
thatbudakguy: I _think_ seo_title is always set (it was set every time I checked), but that order looks right regardless. thanks!

cdhweb/blog/templates/blog/blogpost_detail.html
```diff
@@ -25,24 +23,24 @@
 {% block main %}
 <article typeof="schema:BlogPosting" class="blogpost">
     <header>
-        <h1 property="schema:headline">{{ post.title }}</h1>
-        <meta property="schema:url schema:mainEntityOfPage" content="{% pageurl post %}"/>
+        <h1 property="schema:headline">{{ page.title }}</h1>
+        <meta property="schema:url schema:mainEntityOfPage" content="{% pageurl page %}"/>
```
rlskoeser, line 27: does pageurl return a full url? or can we use `get_full_url` here like you recommended for the site url in the footer? in my experience stuff this kind of metadata does require full urls

cdhweb/pages/templates/cdhpages/home_page.html
```diff
@@ -1,7 +1,8 @@
 {% extends 'base.html' %}
 {% load wagtailcore_tags %}
 
-{% block page-title %}{{ page.seo_title }}{% endblock %}
+{# Override entire title so there's no extra context appended #}
+{% block title %}<title>{% firstof page.seo_title page.title %}</title>{% endblock %}
```
rlskoeser, line 5: 👍

rlskoeser (approved):
looks good!

cdhweb/blog/templates/blog/blogpost_detail.html
thatbudakguy: it's supposed to "intelligently" decide when to return a full url, but probably better to be safe

rlskoeser: I think the "intelligence" is for cross-site / multisite linking; I doubt it could tell that's needed here

PR #302, opened 2021-03-24
clean up mezzanine urls

rlskoeser:
- remove old site index view + template + url, update templates that referenced it
- move error page templates into standard django location
- disable a bunch of mezzanine middleware + context processors


cdhweb/pages/templates/cdhpages/landing_page.html
```diff
@@ -23,7 +23,7 @@
 <header {% if page.header_image %}style="background-image:url('{{ header_img.url }}')"
     {% else %}class="no-background"{% endif %}>
     <div>
-    <a href="{% url 'home' %}" class="home">/</a><h1>{{ page.title }}</h1>
+    <a href="{% pageurl site.root_page %}" class="home">/</a><h1>{{ page.title }}</h1>
```
thatbudakguy, line 26: wow, glad this worked!

templates/base.html
```diff
@@ -1,4 +1,4 @@
-{% load static pages_tags mezzanine_tags fullurl wagtailuserbar %}
+{% load static pages_tags compress wagtailuserbar wagtailcore_tags %}
```
thatbudakguy, line 1: i tested locally and iirc this worked without `pages_tags` which is mezzanine, fyi

templates/base.html
```diff
@@ -102,7 +102,7 @@
 
                     <meta property="schema:alternateName" content="Center for Digital Humanities @ Princeton University" />
                     <meta property="schema:alternateName" content="CDH @ Princeton University" />
-                    <meta property="schema:url" content="{% fullurl 'home' %}"/>
+                    <meta property="schema:url" content="https://{{ site.hostname }}{% pageurl site.root_page %}"/>
```
thatbudakguy, line 106: wonder if it's worth also putting this in the global context processor? not sure where else it'd get used. it also just occurred to me you could try `site.root_page.get_full_url` instead of concatenating with the hostname

thatbudakguy (approved):
everything looks good!

templates/base.html
thatbudakguy: update: tried the latter and it works! it gave me an http URL in local dev, but I think it's smart enough to use https in qa/prod because wagtail takes that into account when constructing URLs. but we can check.

rlskoeser: perfect! will revise, and glad to know about `get_full_url`

cdhweb/pages/templates/cdhpages/home_page.html
```diff
@@ -44,5 +44,8 @@ <h2><a href="{% slugurl 'projects' %}">Projects</a></h2>
 {% endfor %}
 </section>
 
+{% endblock %}
 
-{% endblock %}
\ No newline at end of file
+{% block child-pages %}
+    {% include 'snippets/child_pages.html' with child_pages=page.get_children.live content_only=1 %}
+{% endblock %}
```
thatbudakguy, line 51: nice, this seems like a reasonable solution

thatbudakguy (approved):
let's merge it!

PR #303, opened 2021-03-25
Sitemaps for non-wagtail based views

rlskoeser:
- configure sitemap index and sitemap sections
- create and bind sitemaps for non-page views in people, projects, events apps
- cleaned up some of the old unused sitemap code (I don't think I got it all yet, though)
- sitemap for blog list views
- fix lastmodified list mixin sort logic
- fix people and project list page exodus order to preserve existing nav order (didn't mean to check this one in on this branch! but I did)


cdhweb/events/sitemaps.py
```diff
@@ -1,6 +1,38 @@
...
 
+from cdhweb.events import views
 
-class EventSitemap(PublishedItemSitemap):
-    model = Event
\ No newline at end of file
+
+class EventViewsSitemap(Sitemap):
+    '''Sitemap for project views that are neither Wagtail pages nor
+    tied to models (currently all list views).'''
```
thatbudakguy, line 10: presumably should say "event views"

cdhweb/events/sitemaps.py
```diff
@@ -1,6 +1,38 @@
...
+                url, view
+            ))
+        return urls
+
+    def location(self, obj):
+        return obj[0]
+
+    def lastmod(self, obj):
+        # both pages are modified based on changes to digitized works,
+        # so return the most recent modification time of any of them
```
thatbudakguy, line 37: no "digitized works" here

cdhweb/pages/management/commands/exodus.py
```diff
@@ -243,7 +243,7 @@ def handle(self, *args, **options):
 
             # project list pages become link pages
             project_pages = MezzaninePage.objects \
-                .filter(Q(slug__startswith="projects")).order_by('-slug')
+                .filter(Q(slug__startswith="projects")).order_by('_order')
```
thatbudakguy, line 246: does this need `Q`?

cdhweb/people/sitemaps.py
```diff
@@ -1,11 +1,27 @@
...
+            ('exec-committee', views.ExecListView)
+        ]
+
+    def location(self, obj):
+        # generate url based on archives url names
+        return reverse('people:%s' % obj[0])
+
+    def lastmod(self, obj):
+        # both pages are modified based on changes to digitized works,
+        # so return the most recent modification time of any of them
```
thatbudakguy, line 26: more copypasta'd comments

thatbudakguy (approved):
everything looks good, just some copy-pasted stuff that doesn't match (but we can merge it without addressing that, idc).

sorry for my reliably backwards last-modified code!!

cdhweb/pages/management/commands/exodus.py
rlskoeser: No! Bet it was copied from a different one, that does

cdhweb/people/sitemaps.py
```diff
@@ -1,11 +1,26 @@
...
     def items(self):
-        # only published profiles with is_staff actually have detail pages
-        # and should appear in the sitemap
-        return super().items().filter(is_staff=True)
+        # return list of tuple with url name, view object
+        return [
+            ('staff', views.StaffListView),
+            ('students', views.StudentListView),
+            ('affiliates', views.AffiliateListView),
+            ('exec-committee', views.ExecListView)
```
thatbudakguy, line 17: is it possible to import these from the urlconf so they're not hardcoded? or does that not make any sense

thatbudakguy (approved):
everything looks good, just one general question about importing entries from the urlconf since it seems like the sitemaps are sort of duplicating them. in general, the sitemaps seem to basically reuse logic from a bunch of places (views, urlconfs) so i think if we can just import the relevant stuff that's ideal.

cdhweb/people/sitemaps.py
rlskoeser: I'm not sure how to import just the ones we care about. Did you have an idea about that? 

Agree that there is some duplication here, but if the views change I think we'd likely need to revisit the sitemaps as well.

After writing the blog list sitemap, I wonder if these could be implemented based more directly on the db models, but I don't want to revisit right now since it's at least working this way!

PR #307, opened 2021-03-29
Fix active nav status

rlskoeser:
This addresses setting nav headings to active nav status based on current page being in the page hierarchy. It assumes hierarchy == page url, which I think under wagtail it does.

Suspect that at this point we should go ahead and treat this as a bug and include the fix in 3.0.1 vs trying to include in 3.0...


thatbudakguy (approved):
looks good; templatetag is a reasonable solution for this and i think it should work in all cases.

thatbudakguy:
fwiw I think we should still include in 3.0; any reason not to?

PR #308, opened 2021-03-29
head meta 2

thatbudakguy:
- Use consistent context naming for page meta titles; remove defaults
- Reorder page title calculation logic in base template
- Add page intros as descriptions where present

this fixes issues with list views not having titles/descriptions defined. i made the views consistent across apps; there's no more default for view names (i think you're right that a default is not meaningful) and all views now use the context var `page_title` for the title you see in the h1, the dropdown selector, and the page's metadata.

everything is working except I can't get `striptags` to work on page intros; suspect it's because the object isn't playing nice with django templates. if I apply the filter in the `firstof`, it becomes the string `None` when not present.

### DIFF
--- a/cdhweb/blog/templates/blog/blogpost_archive.html
+++ b/cdhweb/blog/templates/blog/blogpost_archive.html
@@ -1,15 +1,15 @@
 {% extends 'blog/base.html' %}
 {% load humanize %}
 
-{% block page-title %}{{ title|default:'Latest' }} Updates{% endblock %}
+{% block page-title %}{{ page_title }}{% endblock %}
 
 {% block main %}
 <section class="blogposts">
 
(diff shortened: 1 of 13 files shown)


rlskoeser (commented):
Page title updates look good 

I see what you mean about the page intro paragraph! I see the same behavior. If you strip tags or mark safe you get the full content, but then you get None when it isn't set. The problem is that django is autoescaping — https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#firstof  

I did notice when I went looking that the page preview description mixin class has a description field (borrowed from PPA, I think) that — as far as I can tell — isn't actually made available anywhere on the edit forms. I wonder if we should drop it, since we don't need formatted preview functionality anywhere (AFAIK).

Issue #285, opened 2021-02-24
fix temporary models and model fields


thatbudakguy:
several model fields and settings are in a temporary state that needs resolving in 3.0.1.
# People
- [x] `OldProfile` should be removed
# Projects
- [x] `OldProject` should be removed
- [x] `Grant.project` should use the `models.CASCADE` deletion handler and not allow null
- [x] `Grant.old_project` should be removed
- [x] `Membership.project` should use the `models.CASCADE` deletion handler and not allow null
- [x] `Membership.old_project` should be removed
- [x] `ProjectRelatedLink.project` should use the `models.CASCADE` deletion handler and not allow null
- [x] `ProjectRelatedLink.old_project` should be removed
# Events
- [x] `OldEvent` should be removed
# Blog
- [x] `OldBlogPost` should be removed
# Resources
- [x] `DateRange` should be moved to pages app
- ~delete everything else, including the entire app~

rlskoeser:
cleanup migration ran in qa without problems and did not introduce any data errors

PR #310, opened 2021-03-31
Remove old models

rlskoeser:
Have completed the checklist as listed in #285 (including moving and cleaning up a few other things in the resources app) with one exception — I don't think we can delete the resources app yet, because we want to run the migration to clean up the old tables for models in that app.

Removing the old page/pa11y fixtures did prompt me to wonder where we are with automated accessibility testing. Do we have an issue in the backlog somewhere for that?

I propose that removing the empty resources app be included in 3.0.2


cdhweb/context_processors.py
```diff
@@ -1,7 +1,7 @@
 from django.conf import settings
 from wagtail.core.models import Site
 
-from cdhweb.resources.utils import absolutize_url
+from cdhweb.pages.utils import absolutize_url
```
thatbudakguy, line 4: can we check if this is still used anywhere? I think wagtail's `Page.get_full_url()` replaces it in most instances

cdhweb/pages/admin.py
```diff
@@ -7,4 +9,20 @@
...
+
+    def group_names(self, obj):
+        '''custom property to display group membership'''
+        if obj.groups.exists():
+            return ', '.join(g.name for g in obj.groups.all())
+    group_names.short_description = 'groups'
+
+
+admin.site.unregister(User)
+admin.site.register(User, LocalUserAdmin)
```
thatbudakguy, line 28: why do this for django admin when it's available in wagtail admin?

cdhweb/context_processors.py
thatbudakguy: i can't read. this import proves that it is still used

thatbudakguy (approved):
looks good; agree removing resources can be 3.0.2.

re: a11y testing, we have #122, which should probably be rewritten to use Lighthouse but can serve to track it. i think that should be in 3.1 if it's not already.

cdhweb/pages/admin.py
rlskoeser: I copied an existing local admin from another project because I've found it valuable to have an easy way to check user status on the list view. If we can incorporate all of these fields on the wagtail admin then it can go away.

cdhweb/context_processors.py
rlskoeser: yeah, it's still used in a few places for non-page content.  — I wondered if there might be something in wagtail that obviated it, but thought it was ok to leave for now

cdhweb/pages/admin.py
thatbudakguy: I think all we'd need to do is port `group_names`; everything else is already represented there out-of-the-box.

rlskoeser: do you want me to port the differences and remove this?

thatbudakguy: idk how easy it is to customize wagtail admin for built-in stuff like `User`, but you could try doing it the same way we do other models in `wagtail_hooks.py` and see what happens? if it doesn't work, fine to just leave this in.

rlskoeser: I can't find where the existing `ModelAdmin` for `User` is defined! I can add a new one, and found the `add_to_settings_menu` option, but then we have two. I think it would be nice to customize this at some point, but doesn't seem urgent that we do it now.

I did find this https://github.com/wagtail/wagtail/issues/4513 — so maybe User admin isn't a normal model admin right now.

PR #311, opened 2021-04-01
Remove exodus code

rlskoeser:
Removes exodus code; moves `to_streamfield_safe` test utility method into pages conftest

I did a little bit of mezzanine removal, but after running into some errors I think it would be easier to just keep that for 3.0.2 when we can remove it entirely.  (Can't uninstall it because it is referenced in migrations.)

I noticed that the django admin site is throwing an exception — thought it was a change I made with the mezzanine stuff, but turns out we're getting a 500 error. Presumably not urgent at this point since nearly everything can be managed through wagtail now, but makes me think we should try to get 3.0.2 out soon. I think it will be fixed when we remove grappelli, since it's a grappelli template error referencing something in mezzanine that must already have been removed or disabled.


cdhweb/people/views.py
```diff
@@ -239,7 +239,8 @@ def get_context_data(self):
 def speakerlist_gone(request):
     # return 410 gone for speakers list view;
     # (removed in 3.0, no longer needed after the Year of Data)
-    return render(request, 'errors/404.html', context={
+    return render(request, '404.html', context={
         'error_code': 410,
-        'message': "That page isn't here anymore."
+        'message': "That page isn't here anymore.",
+        'page_title': 'Erorr — no longer available'
```
thatbudakguy, line 245: This is a cool way to spell error!

thatbudakguy (approved):

thatbudakguy (approved):

PR #317, opened 2021-04-06
Chore/cleanup migrated content

rlskoeser:
- configures migrated text field to allow links, images embeds
- new one-time use manage command to clean up exodized rich text content to remove unwanted whitespace around inline tags, remove wrapping html/body tags, etc.

### DIFF
--- a/cdhweb/pages/management/commands/cleanup_migrated_content.py
+++ b/cdhweb/pages/management/commands/cleanup_migrated_content.py
@@ -77,7 +77,7 @@ class Command(BaseCommand):
                                 new_revision.publish()
 
                     if 'style="' in cleaned_html:
-                        inline_styles.append(page.get_full_url())
+                        inline_styles.append(realpage.get_full_url())
 
         if inline_styles:
             print('\nPages with inline styles:\n')
--- a/cdhweb/pages/management/commands/cleanup_migrated_content.py
+++ b/cdhweb/pages/management/commands/cleanup_migrated_content.py
@@ -8,6 +8,8 @@ class Command(BaseCommand):
 
     # list of regex patterns to replace with a single space
     re_to_space = [
+        # non-breaking space
+        re.compile(r'&nbsp;'),
         # two or more whitespaces
         re.compile(r'\s\s+'),
     ]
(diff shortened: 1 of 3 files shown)


rlskoeser:
@thatbudakguy I don't know if the list of tags from mezzanine actually does anything, since they don't actually (I think?) correspond to wagtail editor "features".

cdhweb/pages/management/commands/cleanup_migrated_content.py
```diff
@@ -0,0 +1,99 @@
...
+        re.compile(r'\s*</body>\s*</html>$'),
+        # empty div
+        re.compile(r'[\n ]+<div>\s*</div>'),
+    ]
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--noact", action="store_true", default=False,
+            help="Don't save any changes to the database"
+        )
```
thatbudakguy, line 47: love this!! very cool

cdhweb/pages/management/commands/cleanup_migrated_content.py
```diff
@@ -0,0 +1,99 @@
...
+        noact = options['noact']
+        inline_styles = []
+        for page in Page.objects.all():
+            realpage = page.get_specific()
+            # skip pages with no body content
+            if not hasattr(realpage, 'body'):
+                continue
+            for block in realpage.body:
+                # assuming only one migrated block per page
+                if block.block_type == 'migrated':
```
thatbudakguy, line 61: only place this doesn't hold true that I can think of is the form pages (consult/cosponsor); I think I added the iframes to a separate migrated block after the content. but also, is this really assuming that? doesn't look like it is

thatbudakguy (approved):
looks good!

cdhweb/pages/management/commands/cleanup_migrated_content.py
rlskoeser: 🤔 it's assuming that because I'm saving and publishing the document after cleaning up the one block. I didn't know those had a separate block. Should I set a flag if something needs to be saved and then clean after all looping through all the blocks?

thatbudakguy: imo not worth it; there are only (maybe) two pages that are an exception and they don't have much text

Issue #330, opened 2021-06-04
As a User, I want to see CDH Data Fellows listed on the Affiliates and Students pages so I know who received a fellowship.


thatbudakguy:
# testing notes
**note** - some media hasn't been copied over to the test site, so profile pictures and other files may be missing. that's OK.
- check the [students list page](https://test-web.cdh.princeton.edu/people/students/) and confirm that any data fellows listed show the title "Data Fellow" instead of "Data Fellowship Grant Recipient"
- check the same thing for the [affiliates list page](https://test-web.cdh.princeton.edu/people/affiliates/)
- skim the titles of the other profiles on these pages and confirm that none of them appear to have changed/everything else is still correct

# dev notes
**Is your feature request related to a problem? Please describe.**
The new cohort of CDH Data Fellows need to be entered into the system, including their projects and profiles. The projects should work the same as other projects (and as DCGs worked in the past), but getting the profiles onto the appropriate person list pages is more complicated.

- Data Fellows who are students should appear on the students list page
- Data Fellows who are staff/faculty should appear on the affiliates list page
- In both cases, the title displayed should be "Data Fellow"

For more info, [see slack conversation](https://cdhprinceton.slack.com/archives/C2VHT183A/p1622674157002000).

**Describe the solution you'd like**
Most of the logic to determine who goes on what page lives in the `PersonQuerySet`:
https://github.com/Princeton-CDH/cdh-web/blob/27537f7eeed402c0c4d90284a088cdf01b7f09e9/cdhweb/people/models.py#L44-L190

Ideally, the `affiliates()` and `student_affiliates()` methods would change to incorporate Data Fellows into the queryset.

To change the actual title that's displayed, we need to edit the `display_label()` method on the `PersonListView` subclasses, for example:
https://github.com/Princeton-CDH/cdh-web/blob/27537f7eeed402c0c4d90284a088cdf01b7f09e9/cdhweb/people/views.py#L193-L203

For students this may work without modification, since "Data Fellow" is a title (and is likely their most recent title), but for faculty/staff (above) it almost certainly won't, because we normally only display grant awards on that page, and the Data Fellowship is not a grant.

**Describe alternatives you've considered**
The strict separation between students and affiliates deserves more thought, and the two categories may continue to become blurrier/be combined in the future.

rlskoeser:
Let's do the simple version of this, since we still want to display Data Fellow projects and have a grant history for them.
[...]

PR #334, opened 2021-06-10
Update person list display logic for fellowships (#330)

thatbudakguy:
this turned out to be even simpler than I thought — just tweaked the logic slightly and moved the method up to `PersonListView` so it could be shared across subclasses. i looked at the tests and this logic is actually already tested for the case of the "Faculty Fellowship" grant type, so I don't even think we need any new tests! let me know if you agree.

tried with today's dump of production data to get the fellows, and confirmed that both student and staff data fellows display as expected.

### DIFF
--- a/cdhweb/people/views.py
+++ b/cdhweb/people/views.py
@@ -37,6 +37,15 @@ class PersonListView(ListView, LastModifiedListMixin):
         current = self.get_current()
         return self.object_list.exclude(id__in=current.values('id'))
 
+    def grant_label(self, grant):
+        """Convert a grant into a label with date."""
+        # if a fellowship, display as "X fellow"
+        if "Fellow" in grant.grant_type.grant_type:
+            no_ship = grant.grant_type.grant_type.split('ship', 1)[0]
+            return f"{grant.years} {no_ship}"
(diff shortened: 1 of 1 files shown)


cdhweb/people/views.py
```diff
@@ -37,6 +37,16 @@ def get_past(self):
         current = self.get_current()
         return self.object_list.exclude(id__in=current.values('id'))
 
+    def grant_label(self, grant):
+        """Convert a grant into a label with date."""
+        # if a fellowship, display as "X fellow"
+        if "Fellow" in grant.grant_type.grant_type:
+            no_ship = grant.grant_type.grant_type.split('ship', 1)[0]
+            return f"{grant.years} {no_ship}"
```
rlskoeser, line 45: 😆  🚫 🚢

rlskoeser: I might consolidate into a single return since the grant years + label is consistent for both cases, but I don't feel strongly about that. codeclimate or one of the other checkers might complain about an else after a return (unnecessary)

rlskoeser (approved):
Great! glad it was so simple.

Agree the existing test is sufficient.

Issue #337, opened 2021-06-15
Grant with no end date throws error on affiliates list page


thatbudakguy:
# dev notes
**Describe the bug**
When calculating the labels to display for people on the affiliates page, we use the latest grant on which an affiliate was project director. To calculate this, we need to find the most recent grant for which the person's project director role overlapped with the years of the grant. In some circumstances, however, a grant may not have end date set, since this date isn't required. In this scenario, `person.latest_grant()` will return `None`, even though the project directorship is current, because the grant has not ended yet.

**To reproduce**
1. Update or add a grant to a project with a start date but no end date
2. Ensure that a faculty/staff person is added as project director on the project, with dates that overlap with the grant
3. Visit the affiliates page

**Expected behavior**
The label for the affiliate should still display with text like "2021– Dataset Curation Grant Recipient" since the start date is known.

# testing notes
easiest way to test this is to just check that brigid doherty displays correctly on the affiliates list page on the test site, since her project had a grant with no end date that was causing the error. you can also try removing the end date from another current grant and making sure that the person still displays on the affiliates list page, or following the steps under "to reproduce" above.

also make sure to spot-check other folks on the affiliates list page, since their display labels should be unaffected and still show the correct grant.

cvansant:
Looks good! Also tried removing the end date for Arabindan-Kesson, and the profile still displayed correctly.

PR #338, opened 2021-06-15
Allow grants with no end dates on affiliates list (#337)

thatbudakguy:


rlskoeser (changes requested):
Logic looks right to me. Can you add a quick test case for this scenario?

thatbudakguy:
adding the tests was a bit involved because there weren't any for this function and also the test suite wasn't using pytest fixtures, but I'm glad I spent the time to do it, because I found another bug! I think the logic needed to be AND rather than OR in `person.latest_grant()`; let me know if it makes sense to you.

rlskoeser (approved):
I wondered about that! AND for checking both start and end dates, but OR for start date comparison or null — glad you caught it with the tests.

PR #348, opened 2021-07-15
Integrate editoria11y

rlskoeser:
preliminary integration of editoria11y


rlskoeser:
@thatbudakguy rebased against current `develop` — thanks for letting me know I needed to. 

I think the main concerns to resolve for a preliminary integration are:
- What's the best way to manage / include the editoria11y js and styles? I'm including them locally in a vendor folder, but would rather not have to manage copies, esp. since I think John J was already planning on a new release. Can we help him set things up so we can pull in the latest published version?
- Correctly configure editoria11y to be assertive instead of polite when previewing draft pages

One other small thing is we need to override the placement of the icon so it doesn't conflict with the wagtail user bar. JJ suggested just overriding it in our local css, which I guess should be fine.

templates/snippets/editoria11y.html
```diff
@@ -0,0 +1,13 @@
...
+{% if perms.wagtailadmin.access_admin %} {# use same check as wagtailuserbar #}
+{# jquery included in base template #}
+<link rel="stylesheet" media="screen" href="{% static "vendor/editoria11y-1.0.6/css/editoria11y.css" %}">
+<script src="{% static "vendor/editoria11y-1.0.6/js/editoria11y-prefs.js" %}"></script>
+<script src="{% static "vendor/editoria11y-1.0.6/js/editoria11y-localization.js" %}"%></script>
+{% if request.is_preview %}
+<script>
+ed11yAlertMode = "assertive";  {# not sure this is actually working #}
+</script>
+{% endif %}
```
thatbudakguy, line 11: you could put this logic inside of `editoria11y-prefs.js` and then `{% include %}` it instead of using `{% static %}` so that the logic would work inside the JS file. just an idea

thatbudakguy (approved):

thatbudakguy:
If John doesn't have plans to publish as an npm package, I think the easiest thing would be to switch from local includes to using a [git cdn](https://gitcdn.xyz/). So for example https://gitcdn.xyz/repo/itmaybejj/editoria11y/main/js/editoria11y.js would always point to the `editoria11y.js` that's in `main`.

rlskoeser:
How did I not know about git cdn!?  Will that work for CSS as well? 

Any downsides you can see for that approach?

thatbudakguy:
Yep! any public file on github. Only downside is that it's not "officially supported", no uptime guarantees, etc. They're usually services maintained by small orgs or individuals so can theoretically go down or be turned off. But I don't think that's a big problem in this instance, and there's always a few different providers out there, so we can switch if we need to.

templates/snippets/editoria11y.html
rlskoeser: But then wouldn't we have to the prefs js a template file instead? Can we do that with js? (I'm trying to remember if we have... would it just all be inline, then?)

rlskoeser:
@thatbudakguy is there an easy way to set up the static site published from the repo to include a version we can include? or does it probably already have the files we need? https://itmaybejj.github.io/editoria11y/

thatbudakguy:
we could try it; just a question of whether github pages serves with the right CORS headers (I'm guessing not). the URLs would be like https://itmaybejj.github.io/editoria11y/js/editoria11y.js (which works in a browser)

templates/snippets/editoria11y.html
thatbudakguy: Idk about the file extension, but I've definitely done it at least once before. This approach doesn't work if we always want the latest version though, because we have to keep a static copy to use as a template, so probably not worth it

templates/base.html
```diff
@@ -26,16 +26,14 @@
     {% include 'snippets/head_meta.html' %}
     <meta name="google-site-verification" content="pS2kpksxg6JeC90IvA8BHVsFuK_6b7J_vARVLqqu7ck" />
     {% block extra-head %}{% endblock %}
-
+    <script src="https://code.jquery.com/jquery-3.5.1.min.js"
+        integrity="sha256-9/aliU8dGd2tb6OSsuzixeV4y/faTqgFtohetphbbj0="
+        crossorigin="anonymous"></script>
+    {% include "snippets/editoria11y.html" %} {# include before local styles #}
```
thatbudakguy, line 32: just occurred to me that this might work without even including jQuery, if it's already a dependency of semantic UI (and they might even conflict if it is). did you see what happens if you don't include jQuery?

thatbudakguy (approved):
let's merge it!

templates/base.html
rlskoeser: we already had the jquery include, I just moved where it was so it would be available before editoria11y — I had to move up the editoria11y include so our local css would override it, since the editoria11y styles are already set as important 😕  (also this is cdhweb, not ppa, so no semantic ui 🙂 )

thatbudakguy: 🤦🏻 ignore me

Issue #19, opened 2017-08-15
As a user I want to search across all pages by keyword so I can find specific content that interests me.


bwhicks:
# dev notes

- [x] create separate branch to track #359 and remove/reset code related to it from this branch
- [x] add tests for `SiteSearchView`
- [x] implement search result styles from #364 
- [x] implement search bar styles from #367
- [x] add new snapshots/states to percy for visual testing of search

cvansant:
I searched for myself (https://test-web.cdh.princeton.edu/search/?q=camey&page=1)and here's what I found:

- For the most part, the order in which the search results come up makes sense: 1) my staff profile; 2) blog posts in which I am mentioned;  3) blog posts I authored; 4) other pages that mention me. The main question I have relates to 4). I think perhaps pages that mention me should come before the blog posts I authored. For instance, if I were really searching for me, I would want to see that I was listed on the S&Co project page. There certainly could be other views on this, however.
- most of the pages associated with my name say "last updated: April 21, 2021." Perhaps that was the day of the migration? it is not true that most of them were updated then.
- The blog posts are in chronological order, with the earliest ones listed first. I think it would make more sense if the more recent ones were listed first. I don't know what if anything this has to do with the April 21 situation.

I searched for "charter" (https://test-web.cdh.princeton.edu/search/?q=charter&page=1)
- Here, I got the "project charter" page first, which seems perfect. After that, I am unsure of the logic. There are blog posts (in an order I don't understand), but they are interrupted by the PEMM project page. 

Another thought: I tried searching for "PEMM" (https://test-web.cdh.princeton.edu/search/?q=PEMM). I did not get as many results as I did when I searched for "Princeton Ethiopian Miracles of Mary," even though those are the same thing (https://test-web.cdh.princeton.edu/search/?q=%22Princeton+Ethiopian+miracles+of+Mary%22). Is there a way to make all the "Princeton Ethiopian Miracles of Mary" results come up when a user searches for "PEMM"?  Saying this because it seems like abbreviations aren't unusual on CDH web (NLP, etc.).

thatbudakguy:
thought I would post here to add to camey's testing above. Wagtail docs [claim that](https://docs.wagtail.io/en/stable/topics/search/indexing.html#signal-handlers):
[...]

PR #349, opened 2021-07-21
Proof-of-concept page search (#19)

thatbudakguy:
Keeping this as a draft since it's a proof-of-concept implementation/experiment.

Overall, this was very simple to set up. We're using Wagtail's postgres search backend, which is enabled by adding a module to `installed_apps` and setting one setting in `settings.py`. Page content is already setup to be indexed because we defined `search_fields` on our abstract `BasePage` model (which every page type inherits from) to include the `body` field, which is used everywhere for main content. For the two pages with descriptions (`ContentPage` and `BlogPost`), I added their `description` field to `search_fields` as well. The index can be populated on-demand anytime with `python manage.py update_index`, but generally only needs to be run once since signals are automatically set up by wagtail to reindex when page content changes. The only downside to this backend is that it doesn't support partial matching, which you can see if you make a typo. I think this is a small price to pay, though.

I created a very basic view and template that you can view at `/search` with a keyword search and a select box to filter to different types of content. The filtering is done at the database level by selecting a different initial queryset based on the model, and then calling wagtail's search function to apply the keyword query to that queryset. We get a nice bit of free help from wagtail's `parse_query_string`, which automatically converts queries like `foo bar "baz quux" published:yes` into a phrase query for `baz quux`, an ANDed keyword query for `foo AND bar`, and a `filters` dict mapping `published` to `yes`. Not doing anything with the filters yet, but plenty of room to imagine stuff we could do (pub date, etc.) by setting them in the model's `filter_fields`.

There's a wealth of information in the [Wagtail docs on searching](https://docs.wagtail.io/en/stable/topics/search/searching.html), including more info about complex queries, filters, boosting, etc. Note that Wagtail _does_ support searching arbitrary content/django models, but you still need to base that query off of something (i.e. call it on a single queryset). For now, I think it's fine to have the search only operate on `Page` subtypes — that covers all content we can actually link to from a search result, and automatically includes things like landing pages and list pages because they're still `Page` subtypes. I'm not sure what it would look like to search across people or grants, since their search results couldn't really link to anywhere.

### DIFF
--- a/cdhweb/pages/templates/cdhpages/page_search.html
+++ b/cdhweb/pages/templates/cdhpages/page_search.html
@@ -1,0 +1,27 @@
+{% extends 'base.html' %}
+{% load humanize wagtailcore_tags %}
+
+{% block main %}
+<h1>{{ page_title }}</h1>
+<form action="{% url 'search' %}" method="get">
+    <input type="search" name="q" placeholder="{{ query_placeholder }}" value="{{ active_query }}">
+    <select name="filter">
+        {% for filter in filters %}
+        <option value="{{ filter }}"{% if filter == active_filter %} selected{% endif %}>{{ filter|capfirst }}</option>
+        {% endfor %}
+    </select>
+    <input type="submit" value="Search">
+</form>
+<div>
+    {% if results %}
+    <p>{{ results|length }} result{{ results|pluralize}}</p>
+    <ol>
+    {% for page in results %}
+    <li>{{ page.title }} (<b>{{ page.specific.cached_content_type.name }}</b>) — <i>{{ page.first_published_at|naturalday }}</i></li>
+    {% endfor %}
+    </ol>
+    {% else %}
+    <p>No results found.</p>
+    {% endif %}
+</div>
+{% endblock %}
--- a/cdhweb/settings.py
+++ b/cdhweb/settings.py
@@ -35,11 +35,30 @@ TAGGIT_CASE_INSENSITIVE = True
 # https://docs.wagtail.io/en/latest/reference/settings.html#usage-for-images-documents-and-snippets
 WAGTAIL_USAGE_COUNT_ENABLED = True
 
+# Use Wagtail's postgresql search backend.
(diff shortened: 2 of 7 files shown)


cdhweb/settings.py
```diff
@@ -35,12 +35,31 @@
...
 # enable feature detection in images
 WAGTAILIMAGES_FEATURE_DETECTION_ENABLED = True
 
 # custom document model
 WAGTAILDOCS_DOCUMENT_MODEL = "cdhpages.LocalAttachment"
 
+# custom embed finders
+WAGTAILEMBEDS_FINDERS = [
+    {
+        'class': 'wagtail.embeds.finders.oembed'
```
rlskoeser, line 56: possible conflict with your other PR, beware your merge order

cdhweb/pages/templates/cdhpages/page_search.html
```diff
@@ -0,0 +1,27 @@
+{% extends 'base.html' %}
+{% load humanize wagtailcore_tags %}
+
+{% block main %}
+<h1>{{ page_title }}</h1>
+<form action="{% url 'search' %}" method="get">
```
rlskoeser, line 6: will this be a django form once we move this beyond proof of concept?

rlskoeser (commented):
Looks great for a prototype and seems very promising.

I see what you mean now about partial matching — only full words. Definitely seems ok for a first implementation! 

I suspected we'd be limited to a single queryset — searching across different kinds of content is hard! Agree that searching pages gets us a long ways towards what we want.

It might be interesting to think about searching people or images, but I can't imagine that being the highest priority. (And searching profiles for people is probably sufficient, honestly.)

cdhweb/pages/templates/cdhpages/page_search.html
thatbudakguy: definitely; especially if we later want to get into facets and things like that

thatbudakguy:
closing; will reopen "real" implementation later.

Issue #354, opened 2021-07-26
As a Content Editor, I want existing page links to resolve after I rename or move a page so that I don't have to manually create a redirect.


thatbudakguy:
# testing notes
There are two different behaviors that should automatically trigger a redirect to be created:
1. moving a page. you can test this with any content page (for example, the "about" page) — if you move it to live underneath a different page, a redirect should be created so that you can still access it from the old URL, e.g. `/about` instead of `/engage/about` (if you moved it under "engage").
2. editing a page's slug. if you're editing the page in wagtail and check the "promote" tab, you can change the slug (the bit at the very end of the URL). as with above, you should still be able to access it at the old URL, which will redirect you to the new one.

Note that changing a page's title doesn't automatically change its URL or create any redirects! To do that, you need to edit the slug or move the page (the two things above).

# dev notes
see [wagtail recipe docs](https://docs.wagtail.io/en/stable/reference/pages/model_recipes.html#have-redirects-created-automatically-when-changing-page-slug)

thatbudakguy:
bumping to a 2 since this involved a little more than the provided wagtail recipe

rlskoeser:
> bumping to a 2 since this involved a little more than the provided wagtail recipe

makes sense; do you want to revise the issue/story also ? (rename or move?)

cvansant:
1) This worked--I moved "jobs" from "engage" to research and the "engage" url still worked.
2) Also good! Changed the slug for seed grants from "seed-grants" to "seed-grant."

PR #355, opened 2021-07-26
Implement automatic redirects on slug change (#354)

thatbudakguy:
testing was a little confusing; when using the real test client to submit a POST request to change the slug I was getting errors from django-compressor...not sure if it was a red herring, but there was definitely a 500 error that wasn't surfacing properly. settled for just calling the wagtail hook manually to fake the edit event.

### DIFF
--- a/cdhweb/pages/tests/test_hooks.py
+++ b/cdhweb/pages/tests/test_hooks.py
@@ -1,0 +1,20 @@
+from django.urls import reverse
+from wagtail.contrib.redirects.models import Redirect
+from wagtail.core import hooks
+
+
+class TestHooks:
+    def test_create_redirect_on_slug_change(db, rf, content_page):
+        """redirect should automatically be created on slug change"""
+        assert Redirect.objects.count() == 0
+        # get the registered hook so we can call it manually to simulate edit
+        hook_fn = hooks.get_hooks("before_edit_page")[0]
+        request = rf.post(
+            reverse("wagtailadmin_pages:edit", args=(content_page.id,)),
+            {"slug": "new-slug"},  # fake changing the slug
+        )
+        hook_fn(request=request, page=content_page)
+        # should redirect old slug to current version of page
+        assert Redirect.objects.filter(
+            old_path="/landing/content", redirect_page=content_page
+        ).exists()
--- a/cdhweb/pages/tests/test_hooks.py
+++ b/cdhweb/pages/tests/test_hooks.py
@@ -14,7 +14,7 @@ class TestHooks:
             reverse("wagtailadmin_pages:edit", args=(content_page.pk,)),
             {
                 "title": content_page.title,
-                "slug": "new-title",    # change slug
+                "slug": "new-slug",  # change slug
                 "body-count": "1",
                 "body-0-deleted": "",
                 "body-0-order": "0",
@@ -23,6 +23,26 @@ class TestHooks:
                 "attachments-count": "0",
                 "action-publish": "Publish",
             },
+        )
+        # Redirect should be created pointing from old path to current page
+        assert Redirect.objects.filter(
+            old_path=old_path, redirect_page=content_page
+        ).exists()
+
+    def test_create_redirect_on_move_page(db, admin_client, content_page, homepage):
+        """redirect should automatically be created on page move"""
+        assert Redirect.objects.count() == 0
+        # Content page starts out underneath a landing page; we move it to
+        # live underneath the homepage instead: /landing/content -> /content
+        old_path = content_page.url[:-1]
+        admin_client.post(
+            reverse(
+                "wagtailadmin_pages:move_confirm",
+                kwargs={
+                    "page_to_move_id": content_page.id,
+                    "destination_id": homepage.id,
+                },
+            ),
         )
         # Redirect should be created pointing from old path to current page
         assert Redirect.objects.filter(
--- a/cdhweb/pages/tests/test_hooks.py
+++ b/cdhweb/pages/tests/test_hooks.py
@@ -1,20 +1,30 @@
 from django.urls import reverse
 from wagtail.contrib.redirects.models import Redirect
-from wagtail.core import hooks
 
 
 class TestHooks:
-    def test_create_redirect_on_slug_change(db, rf, content_page):
+    def test_create_redirect_on_slug_change(db, admin_client, content_page):
         """redirect should automatically be created on slug change"""
         assert Redirect.objects.count() == 0
-        # get the registered hook so we can call it manually to simulate edit
-        hook_fn = hooks.get_hooks("before_edit_page")[0]
-        request = rf.post(
-            reverse("wagtailadmin_pages:edit", args=(content_page.id,)),
-            {"slug": "new-slug"},  # fake changing the slug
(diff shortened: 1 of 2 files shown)


cdhweb/pages/tests/test_hooks.py
```diff
@@ -0,0 +1,20 @@
...
+        # get the registered hook so we can call it manually to simulate edit
+        hook_fn = hooks.get_hooks("before_edit_page")[0]
+        request = rf.post(
+            reverse("wagtailadmin_pages:edit", args=(content_page.id,)),
+            {"slug": "new-slug"},  # fake changing the slug
+        )
+        hook_fn(request=request, page=content_page)
+        # should redirect old slug to current version of page
+        assert Redirect.objects.filter(
+            old_path="/landing/content", redirect_page=content_page
```
rlskoeser, line 19: This hard-coded string is confusing without context — I presume it is from the `content_page` fixture. Can you get the slug from the object before you change it?

cdhweb/pages/wagtail_hooks.py
```diff
@@ -9,12 +9,23 @@ def global_admin_css():
...
 @hooks.register("register_log_actions")
 def register_exodus_log_action(actions):
     """Add a custom PageLogEntry action to mark page exodus from Mezzanine."""
-    actions.register_action("cdhweb.exodus", "Exodus",
-                            "Migrated from cdhweb v2")
+    actions.register_action("cdhweb.exodus", "Exodus", "Migrated from cdhweb v2")
+
+
+@hooks.register("before_edit_page")
+def create_redirect_on_slug_change(request, page):
```
rlskoeser, line 23: Please add a comment here linking to the wagtail recipe you used and note whether it was copied exactly or adapted in any way.

rlskoeser (approved):
Testing the hook directly seems ok to me, as long as we're confident we're calling it accurately. It would be nice to be able to test it more semantically (it seems like it could come up again). When you were testing with the client, did you use the admin client that is logged in and has permission to edit files?

I see in the wagtail recipe that there are some caveats about this (which make sense); do we need to add anything to our content editor documentation or cdhweb handbook to warn people so they know about it and how to work around it?

thatbudakguy:
I had been using the `admin_client` fixture, but fixing the cascading 500 errors on #260 allowed me to finally debug the test properly, so I was able to rewrite it the "real" way (making a true POST request to edit the page's slug). Added some notes about that and also a reference to the wagtail docs on the actual hook.

Re: the caveats, I'm not sure they matter, but I'm also having trouble understanding what the docs mean when they say:
> Note: This does not work in some cases e.g. when you redirect a page, create a new page in that url and then move the new one.

Could this ever even happen, given that wagtail will block you from creating a page with the same slug as another one? Maybe I'm not totally understanding what they're describing, but it doesn't seem realistic.

cdhweb/pages/tests/test_hooks.py
```diff
@@ -0,0 +1,30 @@
...
+                "slug": "new-title",    # change slug
+                "body-count": "1",
+                "body-0-deleted": "",
+                "body-0-order": "0",
+                "body-0-type": "text",
+                "body-0-value": content_page.body[0].value,
+                "attachments-count": "0",
+                "action-publish": "Publish",
+            },
+        )
```
rlskoeser, line 26: nice

cdhweb/pages/wagtail_hooks.py
```diff
@@ -9,12 +9,25 @@ def global_admin_css():
...
 @hooks.register("register_log_actions")
 def register_exodus_log_action(actions):
     """Add a custom PageLogEntry action to mark page exodus from Mezzanine."""
-    actions.register_action("cdhweb.exodus", "Exodus",
-                            "Migrated from cdhweb v2")
+    actions.register_action("cdhweb.exodus", "Exodus", "Migrated from cdhweb v2")
+
+
+# Copied directly from Wagtail Recipe:
+# https://docs.wagtail.io/en/stable/reference/pages/model_recipes.html#have-redirects-created-automatically-when-changing-page-slug
```
rlskoeser, line 23: thank you!

rlskoeser:
Glad you figured out the 500 error problem! and cool to see the admin client post working and everything doing what you expect.

> Could this ever even happen, given that wagtail will block you from creating a page with the same slug as another one? Maybe I'm not totally understanding what they're describing, but it doesn't seem realistic.

I think it's pretty unlikely, I just think it might be hard for someone to figure out what's going wrong if it does happen and we don't have any documentation. I guess what I imagined is not exactly what they describe, though. The scenario I was thinking of was this: you create a page, decide it's in the wrong place or has the wrong title so you change it, the new hook automatically creates a redirect from the old url to the new one; then, sometime after that, you or someone else creates a new page at the old url. Does the redirect take precedence over the new page? (Maybe it's processed later, in which case maybe it doesn't matter?) It's easy to fix if you know what's happening — just remove the autogenerated redirect.

I think we should document somewhere that redirects are automatically created if you move a page, and maybe we could also think about educating or encouraging people to pay a little more attention to the urls for pages they create (some of our blog post urls are pretty ridiculous).

thatbudakguy:
ah, thanks for explaining. that scenario does make sense. where do you think would be good to document it?

one thing i did just check is that wagtail doesn't automatically change the slug if you change the title. it also doesn't take effect if a page is moved, because its slug doesn't change (wagtail's concept of a slug is only the final portion of the URL, not the full path after the domain like in mezzanine). so, this recipe for automatic redirect behavior only takes effect if someone intentionally changes a page's slug in the admin (which seems like a rare behavior).

looking at the code, i think it might not be too hard to adapt it so that it'd also work when the page is moved — I think we just test if `page.get_url()` has changed (instead of checking the slug). i could try that.

rlskoeser:
oh, interesting; glad you checked — I think it's more likely that people will move pages around in the hierarchy than that they will edit the slug, don't you? I like what you suggest, if it isn't too much more complicated.

thatbudakguy:
OK, I added a second handler for the `before_page_move` hook that also creates a redirect when pages are moved and tested using the same strategy.

cdhweb/pages/tests/test_hooks.py
```diff
@@ -28,3 +28,23 @@ def test_create_redirect_on_slug_change(db, admin_client, content_page):
         assert Redirect.objects.filter(
             old_path=old_path, redirect_page=content_page
         ).exists()
+
+    def test_create_redirect_on_move_page(db, admin_client, content_page, homepage):
+        """redirect should automatically be created on page move"""
+        assert Redirect.objects.count() == 0
+        # Content page starts out underneath a landing page; we move it to
+        # live underneath the homepage instead: /landing/content -> /content
+        old_path = content_page.url[:-1]
```
rlskoeser, line 37: Is the `[:1]` for the trailing slash? seems weird that that's part of the url but not part of the redirect (although I have noticed that when I've used the redirects manually)

rlskoeser (approved):

cdhweb/pages/tests/test_hooks.py
thatbudakguy: I think that's right, but it's weird to me too. that's how the wagtail recipe has it 🤷🏻

Issue #305, opened 2021-03-26
Support embedding Media Central content via oEmbed


thatbudakguy:
# testing notes
You should now be able to embed videos from Media Central in the same way that you'd embed something from YouTube: copy the URL into Wagtail's "embed" block. Any video on Media Central should work; you want to copy the URL that's in the URL bar when you view it. For example: https://mediacentral.princeton.edu/media/Center%20for%20Digital%20Humanities%20Overview/0_mnz807tm

You can test #273 as part of this issue too, since embedding a video using the new method should make it resize properly on mobile devices. After you add a video to the page, try resizing your screen to make it smaller and watch the video resize appropriately.

Note that this won't automatically fix the videos that are currently embedded on the site. For that, we have [an Asana task](https://app.asana.com/0/1134940917892680/1200660870137563/f) to track the work that's needed.

# dev notes
**Is your feature request related to a problem? Please describe.**
The homepage has a prominently displayed video hosted on Media Central that currently can't be represented in Wagtail because there's no embeddable version of that video, so it's stored as "migrated" html with an iFrame and will break if edited in Wagtail.

**Describe the solution you'd like**
Write a custom embed handler for content hosted on [MediaSpace](https://sru.mediaspace.kaltura.com/) that uses the Kaltura player.

**Additional context**
Some docs that might be helpful: https://knowledge.kaltura.com/help/mediaspace-oembed-integration

thatbudakguy:
[list of iframes that need migrating](https://cdhprinceton.slack.com/files/U2VGMNE8P/F01SRDT9TJP/cdhweb-iframes.txt), most of which are media central videos

thatbudakguy:
shifting to a 1 since this turned out to not need tests, and only a bit of regex writing/checking.

cvansant:
1) When I embedded the video, the url in the bar in my browser didn't work (error messages: "The page could not be saved due to validation errors" and "Cannot find an embed for this URL"). Rather, it was the url in the "oEmbed" bar on Media Centralthat worked. Maybe that's what's intended, but wanted to check!

2) Video is resizing appropriately as I make my browser box (is that a term?) smaller and larger!

thatbudakguy:
@cvansant hmm, thanks for testing. can you share the URL that you tried? both the oEmbed URL and the regular one from the URL bar should work (although only on the video's detail page).

cvansant:
[...]

PR #357, opened 2021-07-26
Add Media Central oEmbed provider (#305)

### LINKED ISSUE #273: Embed sizing ignores container margins on mobile
**Describe the bug**
Embedded content, such as videos, is able to escape the width of the page and cause it to have a horizontal scrollbar.

**To reproduce**
1. Visit the site homepage on a mobile device (<768px width)
2. Scroll down until you reach the embedded video about the cdh
3. Observe that it extends outside the viewport to the right

**Expected behavior**
Embeds should be resized to fit the container width.

**Screenshots**
Screen Shot 2021-01-11 at 2 21 19 PM

**Device information**
tested on FF v85 mobile device emulator on OSX 10.15.7

thatbudakguy:
This turned out to be much easier than expected — the [MediaSpace docs](https://knowledge.kaltura.com/help/mediaspace-oembed-integration) actually specify the two valid patterns for oEmbed URLs as well as the correct endpoint, so I didn't even need to write a custom provider (Wagtail's method of supplying the provider config was sufficient). Tested using both URL patterns:
- https://mediacentral.princeton.edu/media/Center%20for%20Digital%20Humanities%20Overview/0_mnz807tm (what you get if you just copy from the URL bar when viewing the video on media central)
- https://mediacentral.princeton.edu/id/0_mnz807tm?width=608&height=402&playerId=25624581 (what you get if you copy from the share/oEmbed tab on that page)

Both work and give the same result.

Also, this appears to fix #273 all on its own: the embed is an HTML5 `<video>` instead of an `<iframe>`, and it respects the container margins! Tested using the FF mobile device emulator but would appreciate confirmation.

### DIFF
--- a/cdhweb/settings.py
+++ b/cdhweb/settings.py
@@ -306,9 +306,26 @@ CACHES = {
     }
 }
 
+# Support embedding content from Princeton's Media Central (Kaltura MediaSpace)
+# See https://knowledge.kaltura.com/help/mediaspace-oembed-integration
+media_central_provider = {
+    "endpoint": "https://mediacentral.princeton.edu/oembed",
+    "urls": (
+        r"^https://mediacentral\.princeton\.edu/id/(?:\w+)\?width=\d+&height=\d+&playerId=\d+$",
+        r"^https://mediacentral\.princeton\.edu/media/[^/]+/(?:\w+)$"
+    )
+}
+
+# These will be tried in order; we put the Media Central one first so that the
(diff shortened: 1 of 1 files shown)


rlskoeser (approved):
That's fantastic that this was so easy and required no code! (too bad we didn't implement it sooner....  why did we not know this?)

Tested locally embedding the video links you included in the PR. Also confirmed that the GlitchHub provider still works correctly.

PR #371, opened 2021-08-10
 implement simple feature flags & purple mode 

rlskoeser:
see checklist in #370 ; I think most of the work is done, just need to add a variant images & favicon and conditional logic on the feature flags

(I branched from `main` because I was thinking of this as a hotfix — because we want to get it out asap — but maybe there aren't that many changes in develop since the last release and I didn't need to do that)


thatbudakguy (approved):
everything looks good here! nice to have a somewhat generalized solution for feature flags so we can take advantage of it later.

Issue #359, opened 2021-08-03
As a user, I want to access the site search via the main navigation so I can start a search from anywhere.


thatbudakguy:
# testing notes
[Figma Designs](https://www.figma.com/file/7P52LnAjsL24HeeJmPChy2/?node-id=44%3A83)

You should now be able to begin your search from any page!
* click the magnifying glass in the top right to start a search.
* you should immediately be able to type into the search box. 
* hitting enter will send you to the search page (#19).
* clicking the "X" will cancel your search and close the bar.

Test the look and behavior of the search bar on both desktop and mobile and note anything that doesn't make sense.

# dev notes

our inspiration for this UX is the main https://www.princeton.edu/ site and its expanding search bar.
see also #19 

- <s> re-add or adapt command to populate site menu so percy previews are more accurate</s> (postpone)
- <s> post a video of the animation for the search bar opening to the PR </s> (postpone)
- [x] test search bar open/close states via percy

cvansant:
I see the x to close the bar on the homepage, but I don't see it on the search page.

cvansant:
I see the x to close the bar on the homepage, but I don't see it on the search page. One thing I also noticed is that the search bar disappears when you scroll down far enough. I don't know if that's a big deal but it happens.

Issue #360, opened 2021-08-03
As a user, I want to limit my site search to particular kinds of content so that I can more quickly find the page I'm looking for.


thatbudakguy:
see also and depends on #19
partially implemented as part of #349

thatbudakguy:
to implement the actual filtering, see discussion: https://github.com/Princeton-CDH/cdh-web/pull/372#discussion_r686981398

PR #372, opened 2021-08-11
19 site search

thatbudakguy:
This PR implements the site search page (#19). For designs, see:
* [Search Form on Figma](https://www.figma.com/file/7P52LnAjsL24HeeJmPChy2/?node-id=44%3A83)
* [Search Results on Figma](https://www.figma.com/file/7P52LnAjsL24HeeJmPChy2/?node-id=0%3A1)

I tried to interpret the search form and how it would fit into the page as best I can, but I think some more discussion is needed to disambiguate the form on this page from the one in #359, which will appear in the top navigation bar (and is not part of this issue).

Would appreciate advice on things like the placeholder text and "no results found" text. The page also doesn't have a title (or rather, the default title of "Search" is hidden) — should it be visible instead?


cdhweb/blog/models.py
```diff
@@ -92,6 +91,9 @@ class BlogPost(BasePage, ClusterableModel, PagePreviewDescriptionMixin):
     ]
     promote_panels = Page.promote_panels + [FieldPanel("tags")]
 
+    # index description in addition to body content
+    search_fields = BasePage.search_fields + [index.SearchField("description")]
```
rlskoeser, line 95: Any other fields we should include for other models, like project members on projects? Or does that only matter when add page-specific filters?

cdhweb/pages/templates/cdhpages/search.html
```diff
@@ -0,0 +1,41 @@
...
+        <span class="current">
+            Page {{ page_obj.number }} of {{ page_obj.paginator.num_pages }}.
+        </span>
+        {% if page_obj.has_next %}
+            <a href="?{% url_replace 'page' page_obj.next_page_number %}">next &rsaquo;</a>
+            <a href="?{% url_replace 'page' page_obj.paginator.num_pages %}">last &raquo;</a>
+        {% endif %}
+    </span>
+</div>
+{% endif %}
```
rlskoeser, line 40: Can't we use the existing pagination template snippet here (snippets/pagination.html)? It looks like you shouldn't have to change it to much — maybe just adapt your url replace logic from this version.

If we use the same snippet, then pagination will be consistent with what we already have and we can change it in one place when we're ready to revise it.

cdhweb/pages/views.py
```diff
@@ -50,3 +60,49 @@ def last_modified(self):
...
+        "updates": BlogPost,
+        "projects": Project,
+        "events": Event,
+    }
+
+    def get_queryset(self):
+        # NOTE not in UI — choose model to search across; Page for everything
+        # filter = self.request.GET.get("filter", "everything")
+        # model = self.filter_models[filter]
+        model = Page
```
rlskoeser, line 87: It still seems weird to me to set/override the model here when you already have it on the class.

When you add model filtering, does it _need_ to query based on the subclass, or can you something more like a normal queryset filter that looks for the related page type? That approach seems more flexible to me — and would make it possible to allow more than one kind of content.

rlskoeser (approved):
Looks good. A few questions and suggestions, no major concerns.

Not sure if it's in scope for this PR, but looks like it would be easy to add a site search config like we did for PPA; would be great to add at some point. See: https://github.com/Princeton-CDH/ppa-django/blob/main/templates/base.html#L23-L26

cdhweb/pages/templates/cdhpages/search.html
thatbudakguy: yeah — I reimplemented here because originally we were going to update the pagination as part of this issue, but now that we've carved it off into a separate design issue i think it makes sense to revert to the default pagination and then change it later in one place.

cdhweb/pages/templatetags/cdh_tags.py
```diff
@@ -55,3 +55,12 @@ def startswith(text, starts):
...
     return False
+
+
+@register.simple_tag(takes_context=True)
+def url_replace(context, field, value):
+    """Add or replace a single GET parameter in a URL."""
+    # Adapted from https://stackoverflow.com/questions/5755150/altering-one-query-parameter-in-a-url-django
+    params = context["request"].GET.copy()
+    params[field] = value
+    return params.urlencode()
```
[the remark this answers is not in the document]
thatbudakguy: oh good call, it's not. i'll check s&co and then add a test

cdhweb/pages/views.py
thatbudakguy: I hadn't thought of filtering based on related page type! The wagtail docs seem to prefer you to do as much as possible on the queryset (including switching the actual model) before calling `search()`, so I wasn't thinking about doing it that way...but I bet it'd work! will try

cdhweb/blog/models.py
rlskoeser: @thatbudakguy I think you could make a best guess now if you want, for the obvious things (like project team members). If you're not comfortable with that you could draft a list and propose the fields & boosting you suggest, which could be reviewed and implemented later.  It occurs to me that generating a list of project members might not be completely trivial — if it isn't then we should probably track it in a separate issue.

cdhweb/pages/templates/cdhpages/snippets/search_result.html
```diff
@@ -0,0 +1,15 @@
+{% load humanize wagtailcore_tags cdh_tags %}
+{# Single search result on site search page #}
+
+{% spaceless %}
+{% autoescape off %}
+<li class="{{ page.get_url|url_to_icon }}">
```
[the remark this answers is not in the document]
rlskoeser: I don't see anything in our currently available icons that would make sense. Hopefully it won't be so long until we're able to revisit and customize styles by type some, so we won't have this forever.  

If you have any ideas of an icon that you think would work, we could suggest it to Gissoo and add to her list of potential new icons. I wondered about a writing icon or pen, but other things could work (and even within our posts we have a lot of variety, too).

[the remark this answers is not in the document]
rlskoeser: The one icon that I did think was kind of appealing is the seed — but maybe that's just because we don't use that icon very much and I like it. What do you think, could it make sense? (sometimes blog posts are like seedlings for other work...)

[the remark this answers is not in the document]
thatbudakguy: I think I'd rather keep the "doc" vs "seed" until we figure out a better alternative...agree that nothing else seems appropriate

cdhweb/pages/views.py
thatbudakguy: reverted this to just use `self.model` for now since the filtering will happen as part of #360

cdhweb/blog/models.py
thatbudakguy: this turned out to be super easy using wagtail's `index.RelatedFields`. I didn't play with the boosting at all, but did confirm that it was working — for example, the davila.js project page comes up (on second page of results) when searching "koeser" even though your name isn't in the page body. i think we can revisit boosting (at query time) later if we decide we need to.

thatbudakguy:
added OpenSearch stuff from PPA, thx for mentioning that

thatbudakguy:
> Placeholder text could be improved; I think I liked the version you had in the prototype better — didn't you list the kinds of content that were searchable?

fwiw i did too, but it turned out to be too long on mobile 😞

cdhweb/pages/templatetags/cdh_tags.py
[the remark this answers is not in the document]
thatbudakguy: added

Issue #368, opened 2021-08-10
As a content editor, I would like to add an individual both to the staff page and to the executive board page.


cvansant:
I was hoping to add MM both to the staff page and to the exec page, but when I added the position "executive board member" and assigned it the same dates as the "faculty director" position, she disappeared from the staff page. 

Even better, I would like it if she could be listed first on the exec page. Model is the UCHV exec page, where Melissa Lane (faculty director) is first: https://uchv.princeton.edu/people/executive-committee.

cvansant:
Update: I tried to add Brian Kernighan--currently on exec--as faculty director, so that he would appear on the staff page. However, his photo was only appearing on the exec page. I ended up having to delete his exec title in order to get him to appear on the staff page. Another reason it would be great to be able to add people two places. Thanks!

Issue #232, opened 2020-11-05
As an admin, I want to control the student/staff status of profiles based on the positions held by people so I can make sure they show up on the right listing page.


thatbudakguy:
**Is your feature request related to a problem? Please describe.**
There's a current conflict in the logic that aggregates profiles between relying on information on a person's "princeton status" (which is initially populated from Princeton's own directory system but can later be changed) and information coming from the positions they've held at CDH or on a CDH-sponsored grant. This leads to inconsistencies in which profiles are displayed on which listing pages.

**Describe the solution you'd like**
The sole determinant of which page a profile appears on should be the position a person held in their relationship to CDH. We currently hard-code these values into lists like `postdoc_titles` and `student_titles`, but they should be able to be set in the admin interface as part of the data of either the title itself, or the particular position instance (person + title + dates). If people have no relationship to CDH through a position (only through a grant), perhaps we can localize such profiles to the Affiliates page, since that page should list all people who have received grants from CDH (including students, though it doesn't currently).

**Additional context**
For an example of currently hard-coded position -> status mappings, see:
https://github.com/Princeton-CDH/cdh-web/blob/96404f5296d0cd9973d21530c3e540813243bd62/cdhweb/people/models.py#L123-L144

PR #373, opened 2021-08-12
Fix current_position_nonexec filter so it only filters out current exec positions, not past

rlskoeser:
this was preventing acting faculty director with past exec membership from showing up on the staff page — they were listed, but always listed as past because the filter logic was catching on their other positions


cdhweb/people/models.py
```diff
@@ -178,7 +178,8 @@ def project_manager_years(self):
     def _current_position_query(self):
         # query to find a person with a current cdh position
         # person *has* a position and it has no end date or date after today
-        return models.Q(positions__isnull=False) & (
+        today = date.today()
+        return models.Q(positions__start_date__lte=today) & (
             models.Q(positions__end_date__isnull=True)
             | models.Q(positions__end_date__gte=date.today())
         )
```
thatbudakguy, line 185: does this get used elsewhere? seems like it'd affect other things too (and glad we fixed it!)

thatbudakguy (approved):
thanks for figuring this out! does this resolve #368?

cdhweb/people/models.py
rlskoeser: I think it gets used in multiple places, which is why we had a method for it. Yeah, not sure how we missed that one (although it's an edge case for sure)

rlskoeser:
> thanks for figuring this out! does this resolve #368?

No, it doesn't — I think it's a separate bug that we should make an issue for. For #368 we need to include faculty director and any interim faculty directors at the beginning of the exec committee member list, and I'm not sure how to do that with the current queries we have without breaking other things. (Maybe we don't make it part of the current query and just look for faculty directors separately from exec...)

rlskoeser:
@thatbudakguy how do you feel about the test coverage? I don't think this method is tested explicitly anywhere

thatbudakguy:
i noticed that too. i'm fine with merging; doesn't seem like something we need to fix right now — might get revisited or invalidated as part of #232 anyway.

PR #375, opened 2021-08-18
359 site search nav

thatbudakguy:
This PR implements #359.

- Adds context processor for site search
- Adds site search to nav

### DIFF
--- a/cdhweb/pages/management/commands/visual_tests.py
+++ b/cdhweb/pages/management/commands/visual_tests.py
@@ -24,7 +24,7 @@ class Command(BaseCommand):
 
         # homepage
         browser.get("http://localhost:8000/")
-        browser.find_element_by_css_selector("label.open").click()
+        browser.find_element_by_css_selector("#mainmenu label.open").click()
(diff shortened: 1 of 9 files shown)


cdhweb/pages/management/commands/visual_tests.py
```diff
@@ -24,6 +24,8 @@ def take_snapshots(self, browser):
 
         # homepage
         browser.get("http://localhost:8000/")
+        browser.find_element_by_css_selector("#mainmenu label.open").click()
+        browser.find_element_by_css_selector("input[type=search]").send_keys("digital")
         percy_snapshot(browser, "Home")
```
rlskoeser, line 29: Should this be a new snapshot? Don't we need to see home page nav without search active?

rlskoeser (approved):

cdhweb/pages/management/commands/visual_tests.py
thatbudakguy: my thought was that we see it on every other page, so no

PR #377, opened 2021-08-19
366 homepage sections

thatbudakguy:
- Add consultations section background image
- Create featured pages section for homepage
- Extract more mixins to file
- Styling for homepage featured pages section

 would appreciate advice on the logic, which is:
- find the about and consult pages via their slugs, if published and existing
- find the first image with the tag "homepage" to use as the main image (of rebecca)
- add all of these to homepage context, and if both pages were present, render a snippet showing them
- resize hi-res photo to retina size using wagtail's smart cropping so it will focus on rm's face (see note below)
- render the colored overlay on the photo using a `linear-gradient` so it can change for purple mode
- choose a background image (of the icons) in css so it can change for purple mode


thatbudakguy:
currently a draft, since I need help on a few things:

- tests are being weird (need to investigate further)
- i think the background image with the consultation icons uses a different white from the actual cdh background color; the borders of the image are noticeable because of this
- the image of rebecca is too small; we need a larger version that can scale up for retina screens. in testing I used the actual source photo from google drive, resized so it was less than 10M, and uploaded to wagtail. This worked ok; it was a little tricky to get it to center her face, but the results might be acceptable. If not, we can re-crop to make sure it centers correctly:

Screen Shot 2021-08-19 at 5 22 11 PM

we won't be able to take advantage of percy visual review for this — both because we are out of snapshots and because it requires too many things to be set in order to work (the image would need to be uploaded to wagtail and multiple pages created with specific content, which is probably more effort than we need right now).

cdhweb/pages/models.py
```diff
@@ -291,6 +292,15 @@ def get_context(self, request):
...
         context["events"] = Event.objects.live().upcoming()[:3]
 
+        # add "featured pages" with special section: currently about/consult,
+        # don't add them to context if not published
+        # NOTE effectively hardcoding by slug for now; could generalize later
+        context["about"] = ContentPage.objects.live().filter(slug="about").first()
+        context["consult"] = ContentPage.objects.live().filter(slug="consult").first()
+
+        # about page image to be used for featured page section
+        context["featured_img"] = Image.objects.filter(tags__name="homepage").first()
```
rlskoeser, line 302: this seems a little strange to me, and potentially unreliable

cdhweb/pages/models.py
```diff
@@ -291,6 +292,15 @@ def get_context(self, request):
         # add up to 3 upcoming, published events
         context["events"] = Event.objects.live().upcoming()[:3]
 
+        # add "featured pages" with special section: currently about/consult,
+        # don't add them to context if not published
+        # NOTE effectively hardcoding by slug for now; could generalize later
+        context["about"] = ContentPage.objects.live().filter(slug="about").first()
+        context["consult"] = ContentPage.objects.live().filter(slug="consult").first()
```
rlskoeser, line 299: When I'm setting multiple things in a dictionary in a row, I prefer to use `update()`

cdhweb/pages/templates/cdhpages/home_page.html
```diff
@@ -18,12 +18,17 @@
...
 {# display editable page content; wrapped for formatting reasons #}
-<section>
+<section class="textcontent">
     <div class="richtext">
     {% include_block page.body %}
     </div>
 </section>
 
+{# featured pages (about & consult) — only render if pages are available #}
+{% if about and consult %}
```
rlskoeser, line 28: why not if only one is available? display doesn't handle just one?

cdhweb/pages/models.py
rlskoeser: I think the image should either be included as an asset in `sitemedia` or it should be configured via an image chooser field/panel on the page it is associated with. I only see you using one size in the template, so I think the simplest thing right now is just to put it in the sitemedia folder. When/if we decided it needs to be configurable we can switch it to a featured image field on the page — but that doesn't seem like it's worth spending time on right now.

cdhweb/pages/tests/test_templates.py
```diff
@@ -128,6 +128,40 @@ def test_upcoming_events(self, client, site, homepage, events):
         response = client.get(homepage.relative_url(site))
         assert events["deadline"] not in response.context["events"]
 
+    @pytest.mark.skip("fixme")
+    def test_empty_featured_pages(self, client, site, homepage):
```
rlskoeser, line 132: what's going wrong with the test? 

this approach seems reasonable enough; I wonder if there's a way to simplify and focus on just the methods/templates you worked on

cdhweb/pages/models.py
rlskoeser: It seems ok to me hardcode by slug for now, but definitely worth noting

rlskoeser (commented):
I think the homepage image should be a static asset, not a wagtail image.

When I test locally, all I'm seeing is the bolded links to read more / request a consultation. Is that expected? I know I don't have the image, am I missing other required content?

cdhweb/pages/templates/cdhpages/home_page.html
thatbudakguy: there's no design for that, is there? what would we do in such a case?

cdhweb/pages/tests/test_templates.py
thatbudakguy: feel free to remove the skip and try it and see if you get the error I did — I didn't understand it at all

cdhweb/pages/templates/cdhpages/home_page.html
rlskoeser: I don't know! I guess you could center it, but agree we don't have a design for that and don't need to support it right now

cdhweb/pages/tests/test_templates.py
rlskoeser: I think the problem was the way you were removing the about page from the hierarchy and then trying to add it again after you deleted it. I'm not sure the right way to do that but since you're already testing that it handles both pages not being present, I simplified and used live/published instead of deleting.

thatbudakguy: that makes sense! I'll do that then. thx for checking!

thatbudakguy:
@rlskoeser if you add a description to the about page, you should now be able to see the whole thing, since the image is hardcoded

rlskoeser (approved):
Looks good.

Should there be a deploy note about adding description to the about page, so we don't forget?

thatbudakguy:
yep, good call

PR #387, opened 2021-09-28
Hotfix/384 fix event ordering

kmcelwee:
For future reference, would this even need a PR?


rlskoeser:
@kmcelwee are there any tests for this method? can we confirm the ordering is correct?

for future reference — PR is good to run automated checks, but might not require review

kmcelwee:
@rlskoeser ✅  d55aebc I wrote a test, let me know if there was a better way to configure the fixture.

rlskoeser:
> @rlskoeser ✅ [d55aebc](https://github.com/Princeton-CDH/cdh-web/commit/d55aebc75832dda7a27cb8f0736e10c4b7435a1c) I wrote a test, let me know if there was a better way to configure the fixture.

There are existing event fixtures in the `conftest.py` under `events`, can you use those?

kmcelwee:
@rlskoeser Changing / expanding those fixture ended up causing trouble in ~5 other tests, and the fixes weren't intuitive. This seemed like the easiest solution, since I only needed one extra event. Otherwise there wouldn't be two upcoming events to compare. If you feel strongly about grouping fixtures, I can debug the tests that are affected by the changing fixtures. Let me know!

rlskoeser:
> @rlskoeser Changing / expanding those fixture ended up causing trouble in ~5 other tests, and the fixes weren't intuitive. This seemed like the easiest solution, since I only needed one extra event. Otherwise there wouldn't be two upcoming events to compare. If you feel strongly about grouping fixtures, I can debug the tests that are effected by the changing fixtures. Let me know!

Ah, I see — no, I don't expect you to change the existing fixtures or other tests! At the least you should add a comment about why you did that. It's a little weird to have the fixture here; if it's easy I'd like to see it as a proper fixture in `conftest` named `upcoming_event` or something like that.

kmcelwee:
@rlskoeser All refactored and good to go! Give me a final 👍 and I'll deliver for testing

rlskoeser (approved):

kmcelwee:
@rlskoeser Thank you for fixing percy. The only change seems to be that two project fixtures swapped in order. When following the view back to the method, it looks like order isn't defined on that queryset: https://github.com/Princeton-CDH/cdh-web/blob/1f5471274f80575c7d9f14e8d0807ecf692f7b12/cdhweb/projects/models.py#L26-L35

I can't find what commit Percy is referencing (I was hoping I'd have something more detailed than "a month ago"), but I also can't see anything in recent commits that would cause this

I'm prepared to ignore this, but it concerns me that I don't know what caused the change. Can different versions of Django have different default orders? Could creating an event fixture change the database enough to cause a small change in the projects table?

[Django docs:](https://docs.djangoproject.com/en/3.1/ref/models/options/#ordering)

> If a query doesn’t have an ordering specified, results are returned from the database in an unspecified order. A particular ordering is guaranteed only when ordering by a set of fields that uniquely identify each object in the results. For example, if a name field isn’t unique, ordering by it won’t guarantee objects with the same name always appear in the same order.

kmcelwee:
@rlskoeser Looking further up the stack trace, on `object_list`'s `get_queryset`, I see that we've already decided to explicitly order projects by grant start date and title:

https://github.com/Princeton-CDH/cdh-web/blob/main/cdhweb/projects/models.py#L68-L75

I assume this is preferred over alphabetical order? This means that it **is** ordered (I was mistaken). Given I'm not misunderstanding anything and https://github.com/Princeton-CDH/cdh-web/issues/389 is defunct. What would next steps look like?

rlskoeser:
@kmcelwee thanks for investigating, I see that filter being applied in the view — that's a change from the 2.x logic, I must have missed that decision, and I don't see it noted anywhere in the change log. I agree we can delete #389 (or comment, close as wontfix and remove the bug label). 

Does this give you any insight into the percy build? did the dates in the fixtures change? is the new listing correct?

kmcelwee:
@rlskoeser OMG
SVP's `start_date` is September 1, 2020
Derrida's `start_date` is "one year ago" (`datetime.today() - timedelta(days=365`)

The latest time we worked on this was August. SO. it's working as it should.
I'm going to accept Percy with your go ahead.

rlskoeser:
> @rlskoeser OMG SVP's `start_date` is September 1, 2020 Derrida's `start_date` is "one year ago" (`datetime.today() - timedelta(days=365`)
> 
> The latest time we worked on this was August. SO. it's sorting properly. I'm going to accept Percy with your go ahead.

Wow! I wondered if there were any relative dates in the fixtures! Should have asked. Ok, so the combination of a relative date and non-relative date shifted the order — display is correct for those dates; and order shouldn't change after this, right?

Agree, accept the Percy change, merge, and 🚀 ! Thanks for the sleuthing 🕵️

PR #455, opened 2025-05-23
FED dependency updates, and related fixes

liamjohnston:
FED dependencies update. Thankfully there were minimal related changes required.

The Sass update has triggered some sass warnings (not errors), which I will address in another PR shortly. It'll mean changing basically all the sass files, but in a fairly uniform way. We've had to do this on a lot of other projects recently too, so should be straightforward.


webpack.config.js
```diff
@@ -25,11 +24,11 @@ module.exports = (env, options) => {
 
   // SVG sprites. Each directory will create a spritemap (spritesheet).
   // It's a good idea to have several different spritemaps, especially to split out
-  // larger or one-off SVGs, so that pages that don't use any of the svgs from that
+  // larger or one-off SVGs, so that pages which don't use any of the svgs from a
```
liamjohnston, line 27: too many `that`s

cdhweb/static_src/components/MainNavDesktop/MainNavDesktop.tsx
liamjohnston, line 1: Latest typescript got a bit grumpy about some previously fine code, but the changes needed were minimal.

cameralibre (approved):

PR #461, opened 2025-06-09
[CDH-98] Make the logo bigger on larger viewports

liamjohnston:
**Associated Issue(s):** #

https://springload-nz.atlassian.net/browse/CDH-98

https://github.com/user-attachments/assets/3609c9c5-ab83-4259-bdc3-4e4c1ad5f4ef

### DIFF
--- a/cdhweb/static_src/global/components/main-nav-desktop.scss
+++ b/cdhweb/static_src/global/components/main-nav-desktop.scss
@@ -65,9 +65,7 @@
     // One-off breakpoint window where we want the logo size to be fluid:
     @media (min-width: fn.em(1616)) {
       width: 10vw;
-    }
-    @include mx.xxxl {
-      width: 180px;
+      max-width: 180px;
     }
   }
 }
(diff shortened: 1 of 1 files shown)


cdhweb/static_src/global/components/main-nav-desktop.scss
```diff
@@ -62,6 +62,13 @@
     @include mx.xl {
       width: 160px;
     }
+    // One-off breakpoint window where we want the logo size to be fluid:
+    @media (min-width: fn.em(1616)) {
+      width: 10vw;
```
cameralibre, line 67: would doing `width: 10vw; max-width: 180px;` work, instead of the `xxxl` rule? I can't access the Jira ticket so I don't know exactly what the issue is.

cameralibre (approved):

cdhweb/static_src/global/components/main-nav-desktop.scss
liamjohnston: In theory, yes. Let me check that though...

PR #472, opened 2025-06-30
Update Developer README Documentation

tanhaow:
This PR improves the developer setup documentation by introducing clearer step-by-step instructions.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,7 +45,7 @@ Development Setup
 Choose one of the following setup options:
 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Option 1: Local Development ("bare-metal")
+Option 1: Local Development ("bare-metal" / non docker)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 This option runs the application directly on your local machine.


README.rst
```diff
@@ -38,74 +38,149 @@ This repository uses `git-flow <https://github.com/nvie/gitflow>`_ conventions;
...
-
-Docker (Springload)
+-----------------
+Development Setup
+-----------------
+
+Choose one of the following setup options:
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Option 1: Local Development ("bare-metal")
```
rlskoeser, line 48:
```suggestion
Option 1: Local Development ("bare-metal" / non docker)
```

README.rst
```diff
@@ -38,74 +38,149 @@ This repository uses `git-flow <https://github.com/nvie/gitflow>`_ conventions;
...
+**Setup Steps:**
+
+1. **Set up Python environment:**
+   .. code-block:: bash
+      # Install Python 3.11 (if not already installed)
+      brew install python@3.11  # macOS
+      
+      # Create and activate virtual environment
+      virtualenv cdhweb -p python3.11
+      source cdhweb/bin/activate
```
rlskoeser, line 62: I prefer to use pyenv to manage python versions; pyenv by default will use the `.python-version` file in the main directory, which is currently 3.12.  This file is also used by GitHub Actions to install the correct python version.

I also prefer the builtin `python -m venv` for creating virtualenvs. 

I don't think we need to require everyone to use the same tools, but following the python version seems important since it might cause problems in some cases.

README.rst
```diff
@@ -38,74 +38,149 @@ This repository uses `git-flow <https://github.com/nvie/gitflow>`_ conventions;
...
+      python manage.py collectstatic --noinput
+
+9. **Run the development server:**
+   .. code-block:: bash
+      python manage.py runserver
+
+10. **Visit the site:**
+    - Main site: http://127.0.0.1:8000/
+    - Django admin: http://127.0.0.1:8000/admin/
+    - Wagtail admin: http://127.0.0.1:8000/cms/
```
rlskoeser, line 109: Ah, nice, this is helpful.  Let's update to use `localhost` - from what I remember, LDAP/CAS login won't work with 127.0.0.1 but it does work with localhost.

README.rst
```diff
@@ -38,74 +38,149 @@ This repository uses `git-flow <https://github.com/nvie/gitflow>`_ conventions;
...
+6. **Set up database:**
+   .. code-block:: bash
+      # Create database
+      createdb cdhweb
+      
+      # Run migrations
+      python manage.py migrate
+      
+      # Create superuser (optional)
+      python manage.py createsuperuser
```
rlskoeser, line 92: It's also possible to initialize a princeton netid account with admin permissions ; both of these accounts may be useful.

```
python manage.py createcasuser --admin netid
```

cdhweb/static_src/fonts/README.md
```diff
@@ -1 +1,9 @@
 Place font files in **this** directory.
+
+Drop all six files in this folder:
+  lineto-brown-regular.woff2
+  lineto-brown-regular-italic.woff2
+  lineto-brown-bold.woff2
+  lineto-brown-bold-italic.woff2
+  lineto-brown-black.woff2
+  lineto-brown-black-italic.woff2
```
rlskoeser, line 9: Thanks for documenting which files are expected, that's helpful.

README.rst
```diff
@@ -38,74 +38,149 @@ This repository uses `git-flow <https://github.com/nvie/gitflow>`_ conventions;
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Option 1: Local Development ("bare-metal")
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This option runs the application directly on your local machine.
+
+**Setup Steps:**
+
+1. **Set up Python environment:**
+   .. code-block:: bash
```
rlskoeser, line 56: This code block isn't rendering properly, probably because this file is ReStructured Text rather than Markdown for historical reasons (we include the readme in our sphinx docs for the project and was probably set up before there was markdown support for sphinx).

rlskoeser (changes requested):
Thank you for improving the documentation!

I'm requesting a few small changes:
- update the code block formatting to ReStructured Text format (RST)
- make sure instructions note the .python-version file; you could provide commands for both brew and pyenv or mention that both are options, I don't feel strongly about this
- update the urls in the local development setup to use localhost instead of IP

I don't feel a strong need to look at it again after you make changes; I'd be ok with you merging once you've addressed my comments. But I'd also be happy to take a quick look if you would find it helpful.

tanhaow:
Thanks for the helpful feedback! I've addressed all the comments and will merge the branch.

PR #478, opened 2025-07-11
update no projects message

sarahframe:
**Associated Issue(s):** https://springload-nz.atlassian.net/browse/CDH-119

### Changes in this PR

Update text displayed when there are no projects which match the chosen filters.

### DIFF
--- a/templates/projects/projects_landing_page.html
+++ b/templates/projects/projects_landing_page.html
@@ -84,7 +84,7 @@
                     {% include 'cdhpages/blocks/tile.html' with internal_page=project tile_type='internal_page_tile' has_component_title=True %}
                 {% endfor %}
             {% else %}
-                <p>No projects found.</p>
+                <p class="no-results-message">No current projects match your filters.</p>
             {% endif %}
             
         </div>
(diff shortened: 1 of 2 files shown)


templates/projects/projects_landing_page.html
```diff
@@ -84,7 +84,16 @@ <h2 class="sr-only">Projects list</h2>
                     {% include 'cdhpages/blocks/tile.html' with internal_page=project tile_type='internal_page_tile' has_component_title=True %}
                 {% endfor %}
             {% else %}
-                <p>No projects found.</p>
+                <div class="no-results-container">
+                <p class="no-results-message">No {% if filter_form.current.value %}current {% endif %}projects match your filters.</p>
+                {% if filter_form.current.value %}
+                    <p class="no-results-suggestion">
+                        <a href="?{% for key, value in request.GET.items %}{% if key != 'current' %}{{ key }}={{ value }}{% if not forloop.last %}&{% endif %}{% endif %}{% endfor %}{% if request.GET.items %}&{% endif %}current=False">
```
dases: Does `value` want `urlencode` filter or would that be a double-up of encoding?

sarahframe: I think it's ok because we dont use that on the original request

PR #482, opened 2025-07-28
CDH-62 add category label to blog page

sarahframe:
**Associated Issue(s):** [#](https://springload-nz.atlassian.net/browse/CDH-62)

### Changes in this PR

Adding category label to blog page, similar to project page

image


liamjohnston:
@sarahframe I _think_ we're meant to be pointing our PRs at `develop` FYI.

sarahframe:
> @sarahframe I _think_ we're meant to be pointing our PRs at `develop` FYI.

That's right thanks I've updated

templates/includes/blog_hero.html
```diff
@@ -4,6 +4,7 @@
     
     <div class="blog-hero__text">
         <h1>{{ page.title }}</h1>
+        {% if page.category %}<div class="blog-hero__category-tag tag tag--dark tag--bigger">{{ page.category }}</div>{% endif %}
```
liamjohnston: Would be a little more readable like so:
```suggestion
        {% if page.category %}
            <div class="blog-hero__category-tag tag tag--dark tag--bigger">{{ page.category }}</div>
        {% endif %}
```
Doesn't need to be on one line.

liamjohnston (approved):
Nice one 🙌 
Just check where this PR is pointing. Pretty sure it should be `develop`.

PR #485, opened 2025-07-29
CDH-62 Add category label to event page

sarahframe:
**Associated Issue(s):** https://springload-nz.atlassian.net/browse/CDH-62

### Changes in this PR

Adding category label event details page.

### Notes
_Include any additional notes that will help in the reviewing of this pull request_

Has already been added to blog page in previous PR

image

### DIFF
--- a/cdhweb/static_src/global/components/event-hero.scss
+++ b/cdhweb/static_src/global/components/event-hero.scss
@@ -59,6 +59,8 @@ Event hero. Similar to standard hero. Key differences:
   grid-area: t;
   font-size: fn.rem(16);
   line-height: 1.125;
+  display: flex; // to support visual re-ordering (tag before h1)
+  flex-direction: column;
 
   :where(h1) {
     margin-block-start: var(--title-top-space);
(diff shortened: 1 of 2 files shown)


cdhweb/static_src/global/components/event-hero.scss
```diff
@@ -88,6 +90,12 @@ Event hero. Similar to standard hero. Key differences:
   }
 }
 
+.event-hero__type-tag {
+  justify-self: start;
+  align-self: start;
+  order: -1; // visually before h1
```
cameralibre, line 96: @sarahframe this shouldn't stop you from merging this, but I can't pass up a good opportunity to nerd-snipe @liamjohnston and learn something new myself: would `reading-flow: flex-visual` be useful here? 
https://developer.chrome.com/blog/reading-flow

It feels like it would help from a **keyboard navigation** perspective (if the tag is a focusable element that takes you to a filtered list of articles with that tag, I dunno if it is, it might just be a `<span>`) 
But I don't really understand how `reading-flow` affects screen reader output. If (as I assume) it alters the screen reader output to read the tag first, then would be we undoing the decision to _nest_ the tag under the `h1` semantically?

cameralibre (approved):
👍🏻

PR #486, opened 2025-08-04
Remove outdated templates

tanhaow:
What's changed:
- Removed `cdhweb/blog/templates`
- Removed `blog tests are also outdated`
- Removed blog tests import from `cdhweb/conftest.py`
- Skipped the unit test related to blog posts in `cdhweb/people/tests/test_pages.py`


cdhweb/people/tests/test_pages.py
```diff
@@ -63,6 +63,7 @@ def test_delete_handler(self, people_landing_page):
         person.delete()
         assert Profile.objects.count() == 0
 
+    @pytest.mark.skip("Skipped after blog tests cleanup - blog integration test")
```
rlskoeser, line 66: I am suspicious of this change... The test should either be removed or fixed, not skipped; you've removed a fixture used for the test, so it would never work again the way it did previously. (And would be harder to figure out later when the fixture is gone but no the test.)

However, I am puzzled why it would be working but not impacted by the code you've removed. I looked briefly at the test fixture and I think maybe it's old code that's no longer used by the running application - but that feels like a bigger change than we intended.

rlskoeser (commented):
I thought cleaning up the templates you flagged as unused would be simpler than this, but it feels like we're pulling on a thread and I can't tell if it's going to start unraveling the sweater or it's part of an unused clump of yarn.   It might be better to leave as is.

I do have a niggling memory that the Springload team kept around old versions of some of the models and templates around for the migration and I'm not sure if they ever cleaned up the code after the migration was complete.

Let's put this on hold and I'll check with Carrie and Springload folks.

PR #491, opened 2025-11-06
[CDH-80] add delete renditions management command

sarahframe:
**Associated Issue(s):** https://springload-nz.atlassian.net/browse/CDH-80

### Changes in this PR

New management command which will delete all the old file type renditions.


dases (approved):
Not sure it matters, but the command name is deleting more than png. :)

